PR #32, opened 2018-05-23
Maint: fix minor typo in coxnet docs

pavopax:


sebp:
I usually use square brackets to denote all sets, including half open sets like the one in the exception. Hence, ``]a, b] ≡ (a, b]``.

PR #59, opened 2019-04-12
Add predict_survival_function and predict_cumulative_hazard_function to CoxnetSurvivalAnalysis

### LINKED ISSUE #46: Add predict_survival_function and predict_cumulative_hazard_function to CoxnetSurvivalAnalysis
The two methods are currently only available to `CoxPHSurvivalAnalysis`, and it is desirable to have them available to `CoxnetSurvivalAnalysis` as well.

sebp:
Estimating the baseline cumulative hazard function requires access to estimated coefficients and survival times from the training data. While `predict` allows specifying arbitrary `alpha` values by interpolating coefficients, we can't do this for `predict_survival_function` and `predict_cumulative_hazard_function` without keeping a copy of the training data. Therefore, both functions only accept `alpha` values that have been used during fitting and are stored in `alphas_`. To avoid increasing training time, estimating the baseline cumulative hazard function is disabled by default, set `fit_baseline_model = True` if you need it.

Fixes #46

### DIFF
--- a/sksurv/linear_model/coxnet.py
+++ b/sksurv/linear_model/coxnet.py
@@ -20,6 +20,7 @@ from sklearn.utils.validation import assert_all_finite, check_array, check_is_fi
 
 from ..base import SurvivalAnalysisMixin
 from ..util import check_arrays_survival
+from .coxph import BreslowEstimator
 from ._coxnet import call_fit_coxnet
 
 __all__ = ['CoxnetSurvivalAnalysis']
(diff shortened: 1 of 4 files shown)


jolyon129:
This new feature is awesome!
I'm trying to use this, but I still got this error 'CoxnetSurvivalAnalysis' object has no attribute 'predict_survival_function',  and I'm using the latest version from conda. 

Should I reinstall from resource to get this new feature?

sebp:
It's currently only available when installing from git. Once version 0.8 is released (probably at the end of the week), conda packages will be updated too.

PR #98, opened 2020-03-12
Remove version on osqp, set min version on pandas

GuillaumeDesforges:
## Motivation

I am packaging this python package for Nix. See https://github.com/NixOS/nixpkgs/pull/82410

## Notes

Using those modifications I have successfully built and tested the package using Nix.

Althought it is clear to me that if those restrictions where placed here, it was for a reason.
Please let me know if something was missed.

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -3,7 +3,7 @@ cvxpy >=1.0
 cython >=0.29
 numexpr
 numpy
-osqp !=0.6.0,!=0.6.1
-pandas >=0.21,<0.26
+osqp >0.6.0
+pandas >=0.21
 scipy >=1.0,!=1.3.0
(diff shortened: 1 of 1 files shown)


sebp:
Thanks for your patch.

Version 0.6.0 and 0.6.1 of OSQP need to remain disallowed, because they 
are unusable, see commit 3337e2789d9fc4aa14512ca2d07c1102710689c7

GuillaumeDesforges:
@sebp do you mean that it is an error on the osqp side, another requirement of this package, or something that could be fixed inside this package ?

sebp:
It's an issue in osqp itself, which they have fixed in later versions.

GuillaumeDesforges:
Ok I will update osqp on Nixpkgs then. Thanks

Issue #42, opened 2018-09-06
Different results of CoxPHSurvivalAnalysis and CoxnetSurvivalAnalysis


leihuang:
The documentation of `CoxPHSurvivalAnalysis` says: 
> Cox proportional hazards model.

And the documentation of `CoxnetSurvivalAnalysis` says:
> Cox's proportional hazard's model with elastic net penalty.

So I assume the two classes implement the same model, and should return the same results when set with the same model parameters and given the same data. However, I see different results. Why? Also, what are the differences between them?

Codes:
```python
from sksurv.linear_model import CoxPHSurvivalAnalysis, CoxnetSurvivalAnalysis
from sksurv.datasets import load_veterans_lung_cancer
from sksurv.preprocessing import OneHotEncoder

X_, y = load_veterans_lung_cancer()
X = OneHotEncoder().fit_transform(X_)

# try to match the model parameters wherever possible
f = CoxPHSurvivalAnalysis(alpha=0.5, n_iter=100000)
g = CoxnetSurvivalAnalysis(alphas=[0.5], alpha_min_ratio=1, n_alphas=1, 
                           l1_ratio=1e-16, tol=1e-09, normalize=False)

print(f)
print(g)

f.fit(X, y)
g.fit(X, y)

print(f.coef_)
print(g.coef_[:,0])
```

Output:
```python
CoxPHSurvivalAnalysis(alpha=0.5, n_iter=100000, tol=1e-09, verbose=0)
CoxnetSurvivalAnalysis(alpha_min_ratio=0.0001, alphas=[0.5], copy_X=True,
            l1_ratio=1e-16, max_iter=100000, n_alphas=1, normalize=False,
            penalty_factor=None, tol=1e-09, verbose=False)
[-8.34518623e-03 -7.21105070e-01 -2.80434400e-01 -1.11234345e+00
 -3.26083027e-02 -1.93213436e-04  6.22726190e-02  2.90289950e-01]
[-0.00346722 -0.05117406  0.06044394 -0.16433136 -0.03300373  0.0003172
 -0.00881617  0.06956854]
```

What I've gathered:
* `CoxPHSurvivalAnalysis` is `sksurv`'s own implementation of Cox Proportional Hazard model, and supports ridge (L2) regularization. 
* `CoxnetSurvivalAnalysis` is a wrapper of some C++ extension codes used by R's `glmnet` package, and supports elastic net (L1 and L2) regularization. 
[...]

PR #106, opened 2020-04-11
Allow per-feature regularization for CoxPHSurvivalAnalysis

### LINKED ISSUE #102: Adding penalty_factor to CoxPHSurvivalAnalysis
It would be very useful to also support a `penalty_factor` in `CoxPHSurvivalAnalysis` in order to always include unpenalized covariates in the model. This is important when you need to adjust for e.g. known prognostic clinical or molecular covariates which shouldn't be penalized.  This is something supported for Cox ridge regression in for example the `penalized` R package.

sebp:
Can be used to have features entering the model unpenalized.

Fixes #102

### DIFF
--- a/sksurv/linear_model/coxph.py
+++ b/sksurv/linear_model/coxph.py
@@ -10,14 +10,14 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import numbers
 import warnings
 
 import numpy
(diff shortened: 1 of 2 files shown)


hermidalc:
Very nice work @sebp.  I've written my own `FastCoxPHSurvivalAnalysis` sksurv class which inherits from `CoxnetSurvivalAnalysis` but has fixed settings to force L2 ridge regression based on info in #42.  I will compare the results

PR #134, opened 2020-07-23
Improvement of calculating cumulative dynamic auc/roc

laqua-stack:
**Checklist**

- [ ] py.test passes
- [ ] tests are included
- [x] code is well formatted

Calculation of AUC is faster due to use of numpy routines instead of for loops.
In addition this generalizes the time depending ROC/AUC to time-varying estimates (e.g. predicted survival function from RSF). 
Instead of `(n_samples,)` shaped estimate `(n_samples, n_times,)` are given, where the `estimate` for each timepoint in `times`, the AUC/ROC should be calculated for should be given as input along `axis=1`.
By specifiying `ret_roc=True`, the tpr and fpr for each timepoint in `times` is returned and can be used to plot time-depedent roc curves.
TODOs:
1.) there are minor differences in some of the calculated AUCs from pytest; I assume, that they result from numerical issues or dtype conversions; need to take a closer look. Hot candidates are the scipy integration and the division by the sum of ipcw / sum of controls.
2.) Due to the use of numpy.take_along_axis we would need to update numpy to atleast 1.15. when it was added.
3.) One test is failing, since normally 2 dimensional estimate arrays raise an error, this is not longer necessary, since this behavior is reasonable.
4.) We need new test for the scenario in 3
5.) There are still some issue concerning the Tox Code Format test with a line continuation and indents; which I (shame on me ;-)) couldn't fix.

### DIFF
--- a/sksurv/metrics.py
+++ b/sksurv/metrics.py
@@ -449,7 +449,7 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
             raise TypeError
     estimate=numpy.atleast_1d(estimate)          
     if estimate.ndim == 1:
-        numpy.tile(numpy.expand_dims(estimate,axis=1),(1,n_times))
+        estimate=numpy.tile(numpy.expand_dims(estimate,axis=1),(1,n_times))
     estimate = _check_estimate2D(estimate, test_time)
     
     if times.max() >= test_time.max() or times.min() < test_time.min():
--- a/sksurv/metrics.py
+++ b/sksurv/metrics.py
@@ -37,6 +37,15 @@ def _check_estimate(estimate, test_time):
     return estimate
 
 
+def _check_estimate2D(estimate, test_time):
+    estimate = check_array(estimate, ensure_2d=True)
+    if estimate.ndim != 2:
+        raise ValueError(
+            'Expected 2D array, got {:d}D array instead:\narray={}.\n'.format(
+                estimate.ndim, estimate))
+    check_consistent_length(test_time, estimate)
+    return estimate
+
 def _check_inputs(event_indicator, event_time, estimate):
     check_consistent_length(event_indicator, event_time, estimate)
     event_indicator = check_array(event_indicator, ensure_2d=False)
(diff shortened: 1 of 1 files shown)


sksurv/metrics.py
```diff
@@ -411,65 +421,109 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
...
+                                     )
+                                 )
+                             , tied_tol)
+
+    add_tp =  numpy.multiply(is_case, ipcw)
+    add_fp =  numpy.multiply(is_control, 1)
+
+    cumsum_tp = numpy.cumsum(add_tp, axis=0)
+    cumsum_fp = numpy.cumsum(add_fp, axis=0)
+    true_pos = numpy.divide(cumsum_tp, add_tp.sum(axis=0))
```
sebp, line 500: `add_tp.sum(axis=0)` is the same as `cumsum_tp[-1]`

sksurv/metrics.py
```diff
@@ -411,65 +421,109 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
...
+    except: # list
+        if isinstance(times, list, ):
+            n_times = len(times)
+            times = numpy.array(times)
+        else:
+            raise TypeError
+
+    estimate = numpy.atleast_1d(estimate)          
+    if estimate.ndim == 1:
+        estimate=numpy.tile(numpy.expand_dims(estimate, axis=1), (1, n_times, ))
```
sebp, line 452: Please use numpy.repeat insted of numpy.tile(numpy.expand_dims(…))

sksurv/metrics.py
```diff
@@ -411,65 +421,109 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
...
+                            test_event)
+    is_control = numpy.greater_equal(test_time, times)
+    n_controls = is_control.sum(axis=0, )
+    is_tied = numpy.less_equal(
+                    numpy.absolute(numpy.subtract
+                                     (
+                                             estimate, numpy.roll(estimate, 1, axis=0, )
+                                     )
+                                 )
+                             , tied_tol)
```
sebp, line 493: I don't think this construct works all of the time, because `numpy.roll(estimate, 1, axis=0)` moves that last element of `estimate` into the first position, thus the first element of `is_tied` could be True, whereas this never could be the case with the previous code. A unit test would definitely be required here.

sksurv/metrics.py
```diff
@@ -411,65 +421,109 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
...
+
+    add_tp =  numpy.multiply(is_case, ipcw)
+    add_fp =  numpy.multiply(is_control, 1)
```
sebp, line 496: Multiplication by 1 can be removed.

sksurv/metrics.py
```diff
@@ -411,65 +421,109 @@ def cumulative_dynamic_auc(survival_train, survival_test, estimate, times, tied_
...
            Statistical Methods in Medical Research, 2014.
     """
+    """
+       IMPROVEMENT:
+       numpy calculation for speed up, returns also the IPCW ROC pairs tpr and fpr
+       test_time, test_event and times should be arrays with shapes:
+           test_time:(n_samples,)
+           test_event: (n_samples,)
+           times: (n_times,)
+           estimate:(n_samples,) NOTE: could be generalized to (n_samples,n_times,)
```
sebp, line 431: Support for having distinct estimates for each time point would be great.

amaliaspa:
I have successfully used Random Survival Forests for predictions and evaluated the model with the time-dependent AUC. I need now to generate the ROC curves themselves for different points in time (the same used for the AUC) and I found this issue, though it's not clear for me how to count tpr and fpr. Could you @laqua-stack or @sebp help with this? 
Thanks in advance!

PR #165, opened 2021-01-20
Feat/min impurity decrease

### LINKED ISSUE #144: Consider opening up `min_impurity_decrease` in survival tree for early stoppping?
Hello there, 
Is it correct understanding that `proxy_impurity_improvement` here is in fact the log-rank z stats? In which case is it possible to open up `min_impurity_decrease` as a param like in sklearn as early stopping criteria instead of forcing it to be 0?
https://github.com/sebp/scikit-survival/blob/3f19c9fccbae544e5f3d68d3bc9c7186559c887e/sksurv/tree/tree.py#L227
Thanks,
Agnes

agnesbao:
**Checklist**

- [x] closes #144
- [x] py.test passes
- [x] tests are included
- [x] code is well formatted

Add `min_logrank_split` parameter to `SurvivalTree` which function in place as `min_impurity_decrease` in sklearn tree model. Like in DecisionTree we set min threshold on whether to keep splitting based on impurity decrease, here we set min threshold on the abs logrank stats (how separate the survival curves are)

I'm still working on writing a functional test for this param, but want to open this PR first to seek feedback on whether it makes sense.


sebp (changes requested):
Thanks for helping to improve sksurv. The changes look reasonable. However, you would also need to modify `RandomSurvivalForest` to include the new `min_logrank_split` parameter. In addition, you need to add tests to assure that the intended change does what it is supposed to.

I'd suggest to create some instances where the `min_logrank_split` gets triggered at different depth of the tree building process. You can check the log-rank statistic in a fully grown tree and design tests based on that. You might find the [LogrankTreeBuilder class](https://github.com/sebp/scikit-survival/blob/a03e7e27cb83416206cf76fcc57ab416a9d9e121/tests/test_tree.py#L73) helpful in that regard.

Please let me know if you have any questions.

tests/test_tree.py
```diff
@@ -138,12 +143,14 @@ def _get_best_split(self, X, y):
                 t = (v + values[i + 1]) * 0.5
                 groups = (vals <= t).astype(int)
                 if groups.sum() >= min_leaf and (X.shape[0] - groups.sum()) >= min_leaf:
-                    s, _ = compare_survival(y, groups)
-                    if s > best_stat:
+                    s, p = compare_survival(y, groups)
+                    abs_z = abs(norm.ppf(p/2))
+                    if s > best_stat and abs_z >= min_logrank:
```
sebp, line 148: The `LogrankCriterion` class uses the absolute value of the log-rank test statistic as splitting criterion, whereas here, the condition is based on the p-value.

sebp (changes requested):
First, I think the min log-rank check in the unit test differs from what `LogrankCriterion`.

Second, I'm not sure whether the unit test is sufficient in the sense that this condition is actually triggered.

sebp:
I'm closing this PR as no further information has been provided. Please feel free to reopen this PR if you can provide the requested updates.

PR #203, opened 2021-05-26
DOC: Remove advice to log-transform times prior to SVM training as this is done internally

zsewa:
Hi,
this pull request aims to update the documentation on the Survival SVM. The jupyter notebook is adjusted according to discussion #200.

In the example notebook:
- survival times are no longer log-transformed prior to fitting the regression objective SVM
- reverting the log-transformation on predictions thus is not needed anymore

- The notebook was rerun in order to update the displayed outputs

Cheers!


sebp:
Thanks a lot for helping to improve the documentation!

PR #243, opened 2021-12-14
Fix a variety of build problems.

llpamies:
**Checklist**

- [x] py.test passes

In LLVM, this project was not compiling properly. With these changes, the project seems to compile fine.


sebp:
Thanks a lot for your PR.

Including `coxnet.h` in `coxnet.hxx`, which itself is included in `coxnet.h` seems strange.

llpamies:
You are right, the circular dependency I included is not the right solution. What happens is that `coxnet.hpp`  should not be a header file (since it doesn't contain declarations, but definitions). Basically, coxnet.hpp should be renamed to coxnet.cpp (or .cc, or .c++) and then coxnet.h should be included in the source file.

I a new commit in this PR I renamed coxnet.hpp to coxnet.cc, and removed the circular dependency.

llpamies:
Unfortunately these double (circular) includes are necessary. There are other alternatives such as 1) moving all logic from coxnet.hxx into coxnet.h, or add explicit template declarations inside coxnet.hxx. The list of alternatives is well covered here:
https://www.codeproject.com/Articles/48575/How-to-Define-a-Template-Class-in-a-h-File-and-Imp

What I have done in this PR is described as "Method 3" in the link above. Note that they also have the circular include dependency.

sebp:
Thanks for clarifying. I think the least confusing solution would be to copy `coxnet.hxx` into `coxnet.h` so that only `coxnet.h` remains. In addition, I believe the includes of `cstdint` can be removed, because only `size_t` is required, which is part of `cstddef`.

llpamies:
Thanks @sebp, sorry, the holidays have delayed this.

I merged `coxnet.hxx` into `coxnet.h`, and changed `cstdint` to `cstddef` where it was necessary.

sebp:
Thanks for the update!

I should have better explained what I mean by merging `coxnet.hxx`. I was suggesting to keep the definitions as defined in `coxnet.h` and put the implementations after it by essentially replacing `#include "coxnet.hxx"` with the contents of that file (minus the namespace and includes).

Also, for `isnan` and `isfinite`, please include the `cmath` header.

llpamies:
Done.

sebp:
I fixed CI in master. Could you please rebase upon the latest master?

llpamies:
I think I have done it.

sebp:
Merged with commit c70d3ad31c2825370c00529e718b9822da0ed7b0

PR #271, opened 2022-06-01
fixed bracket direction

JimmyJHickey:

I fixed a bracket that was facing the wrong direction in an error message.


sebp:
Thanks for helping to improve scikit-survival.

The bracket is actually intentional to indicate the the upper limit is excluded, i.e. `times.max() < test_time.max()` should hold.

PR #291, opened 2022-08-22
Add `apply` and `decision_path` to `SurvivalTree`

### LINKED ISSUE #290: `RandomSurvivalForest.apply` raises `AttributeError`
**Describe the bug**

The apply method inherited from scikit-learn is broken. After fitting, apply raises AttributeError since it is not implemented in the SurvivalTree class.

**Code Sample to Reproduce the Bug**

```python
from sksurv.datasets import load_whas500
from sksurv.ensemble.forest import RandomSurvivalForest

X_sk, y_sk = load_whas500()
rf = RandomSurvivalForest()
rf.fit(X_sk, y_sk)
rf.apply(X_sk)
```

**Expected Results**
Same data-structure as returned by scikit-learn RF.

**Actual Results**
```python-traceback
Traceback (most recent call last)
[/usr/local/lib/python3.7/dist-packages/sklearn/ensemble/_forest.py](https://localhost:8080/#) in <genexpr>(.0)
    251             verbose=self.verbose,
    252             **_joblib_parallel_args(prefer="threads"),
--> 253         )(delayed(tree.apply)(X, check_input=False) for tree in self.estimators_)
    254 
    255         return np.array(results).T
AttributeError: 'SurvivalTree' object has no attribute 'apply'
```

**Versions**
Please execute the following snippet and paste the output below.

```python
System:
    python: 3.7.13 (default, Apr 24 2022, 01:04:09)  [GCC 7.5.0]
executable: /usr/bin/python3
   machine: Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic

Python dependencies:
          pip: 21.1.3
   setuptools: 57.4.0
      sklearn: 1.0.2
        numpy: 1.21.6
        scipy: 1.7.3
       Cython: 0.29.32
       pandas: 1.3.5
   matplotlib: 3.2.2
       joblib: 1.1.0
threadpoolctl: 3.1.0

Built with OpenMP: True
sksurv: 0.17.2
cvxopt: 1.3.0
cvxpy: 1.2.1
numexpr: 2.8.3
osqp: 0.6.2
```

Vincent-Maladiere:
**Checklist**

- [x] closes #290
- [x] py.test passes
- [x] tests are included
- [x] code is well formatted

Add `apply` and `decision_path` to `SurvivalForest` to also enable the same methods for `RandomSurvivalForest` and `ExtraSurvivalTrees`.


tests/test_forest.py
```diff
@@ -260,3 +260,20 @@ def test_fit_max_samples(make_whas500, forest_cls, max_samples, exc_type, exc_ms
...
+    forest.fit(whas500.x, whas500.y)
+
+    x_trans = forest.apply(whas500.x)
+
+    assert x_trans.shape[0] == whas500.x.shape[0]
+    assert x_trans.shape[1] == forest.n_estimators
+
+    x_path, _ = forest.decision_path(whas500.x)
+
+    assert x_path.todense().shape[0] == whas500.x.shape[0]
```
ogrisel, line 279: `.todense()` return a `np.matrix` instance which are considered soft deprecated. It's recommended to use `.toarray()` instead.

sebp:
CI is failing, because conda re-build the scipy package yesterday against an incompatible numpy version. I hope they will fix it soon.

Vincent-Maladiere:
Any further suggestions @sebp?

tests/test_forest.py
ogrisel: @Vincent-Maladiere this comment was not addressed :)

Vincent-Maladiere: Yes indeed, done :)

tests/test_tree.py
```diff
@@ -496,3 +496,81 @@ def test_X_idx_sorted(fake_data, val):
...
+    assert_array_equal(X_trans, X_trans_expected)
+
+    X_path = tree.decision_path(X_test).toarray()
+    X_path_expected = numpy.array([
+        [1, 1, 0, 0, 0],
+        [1, 1, 0, 0, 0],
+        [1, 0, 1, 1, 0],
+        [1, 0, 1, 0, 1],
+        [1, 0, 1, 1, 0],
+    ])
```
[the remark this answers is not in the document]
ogrisel: Actually now I realize that the index of values returned by `apply` are node indices for leaf nodes but the numbering is for all nodes (decision non-leaf node included). This is actually explained in the docstring.

So we just have to check that:

```python
assert_array_equal(
    X_path[np.arange(X_test.shape[0]), X_trans], np.ones(X_test.shape[0])
)
```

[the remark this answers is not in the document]
Vincent-Maladiere: Nice catch! I prefer this approach as well

tests/test_tree.py
```diff
@@ -496,3 +496,81 @@ def test_X_idx_sorted(fake_data, val):
...
+def test_apply_values():
+    """Test `apply` and `decision_function` with a toy dataset.
+
+    The duration is directly proportional to x_1 in case of event.
+    In case of censoring, the duration follows a uniform distribution.
+
+    Besides, we observe an event only when `x_1 > 0`.
+    `x_2` follows a uniform distribution and brings zero predictive power.
+
+    Thus, the decision path only depends on the value of x_1.
```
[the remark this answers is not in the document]
Vincent-Maladiere: It is kinda checked by showing that the `apply` and `decision_path` output depends only on `x_1` and not `x_2`. However, I agree this is an arbitrary test that may not provide much value.

sebp (changes requested):
The API  doc mentions that `X` can be a sparse matrix, but currently the API doc of `fit` does not mention sparse matrices nor are there tests for it. I'd suggest to either add tests for sparse matrices or remove it from the API  doc.

Vincent-Maladiere:
Hi @sebp, I have issues with `pytest --nbval`, this test fail on the `master` branch as well. Should we ignore it?

sebp:
It has been fixed in commit 2b3cfa55ef485fc306133f7e5e36415e8cc7a83f, just rebase on top of the master branch.

sebp:
Thanks for your work!

The current tests for sparse matrices cover `fit`, `apply`, and `decision_path`, but not any of the `predict` functions. I would appreciate if you could add these too. I'd suggest to compare the result to the one obtained with a dense matrix.

tests/test_tree.py
```diff
@@ -496,3 +498,89 @@ def test_X_idx_sorted(fake_data, val):
...
+    X_path = tree.decision_path(X_sparse).toarray()
+
+    assert X_path.shape[0] == X.shape[0]
+    assert X_path.shape[1] == tree.tree_.node_count
+
+    ones = X_path[numpy.arange(X.shape[0]), X_trans]
+    assert_array_equal(ones, numpy.ones(X.shape[0]))
+
+
+@pytest.mark.parametrize("seed", range(100))
```
ogrisel, line 549: It's good that you checked that this test passes on many seeds but I think in the long run it's better to only run on a single seed to avoid wasting CI resources.

tests/test_tree.py
```diff
@@ -496,3 +498,89 @@ def test_X_idx_sorted(fake_data, val):
...
+def test_predict_sparse(seed, make_whas500):
+    whas500 = make_whas500(to_numeric=True)
+    X, y = whas500.x, whas500.y
+    # Duplicates values in whas500 leads to assert errors because of
+    # tie resolution during tree fitting.
+    # Using a synthetic dataset resolves this issue.
+    X = numpy.random.randn(*X.shape)
```
ogrisel, line 556: Let's make this test fully deterministic.

```suggestion
    X = numpy.random.RandomState(0).randn(*X.shape)
```

sebp: Using `numpy.random.RandomState.binomial` would be better, because when using `randn` there is likely no difference to a dense matrix.

tests/test_tree.py
```diff
@@ -543,3 +544,43 @@ def test_apply_sparse(veterans):
...
+
+    assert_array_equal(y_pred, y_pred_csr)
+
+    for step_f, step_f_csr in zip(y_pred_cum_h, y_pred_cum_h_csr):
+        assert_array_equal(step_f.x, step_f_csr.x)
+        assert_array_equal(step_f.y, step_f_csr.y)
+
+    for step_f, step_f_csr in zip(y_pred_surv, y_pred_surv_csr):
+        assert_array_equal(step_f.x, step_f_csr.x)
+        assert_array_equal(step_f.y, step_f_csr.y)
```
sebp, line 586: You should also check attributes `a` and `b`, or just implement [`__eq__()`](https://docs.python.org/3/reference/datamodel.html#object.__eq__) in [StepFunction](https://github.com/sebp/scikit-survival/blob/master/sksurv/functions.py)

sebp (changes requested):

sksurv/functions.py
```diff
@@ -75,3 +75,11 @@ def __call__(self, x):
...
     def __repr__(self):
         return "StepFunction(x=%r, y=%r, a=%r, b=%r)" % (self.x, self.y, self.a, self.b)
+
+    def __eq__(self, other):
+        return (
+            all(self.x == other.x)
+            and all(self.y == other.y)
+            and self.a == other.a
+            and self.b == other.b
+        )
```
sebp, line 85: You should add `if isinstance(other, type(self)):` and return `False` otherwise.

sebp:
Thanks for your work! It has been merged in commit 47d7ba6029c1f9cda48817594c8f585451bc472e

sebp:
I just fixed in master, maybe rebase your branch.

PR #362, opened 2023-05-10
reduce concordance_index_censored's memory use

cpoerschke:
**Checklist**

- [X] pytest passes
- [ ] tests are included - please see note and code snippet below
- [X] code is well formatted

issue encountered: Passing relatively lots of technical data samples to `concordance_index_censored` results in "kernel died" in a notebook environment.

## analysis

* The `concordance_index_censored` computation calls [_get_comparable](https://github.com/sebp/scikit-survival/blob/v0.20.0/sksurv/metrics.py#L89-L114) which returns up to `N` masks of length `N` where `N` is the number of samples being scored.
* The `_get_comparable` returns all the masks and the caller iterates over all the masks but it uses only one mask at a time.

## proposed solution

* Change `_get_comparable` to return (smaller) mask information instead of (potentially much larger) full mask content.
* Change the `_get_comparable` caller to construct the full mask from the mask info, one mask at a time at the point of use.

## testing

Existing tests continue to pass. In terms of additional test coverage, I've tried to build a [test_metrics.py](https://github.com/sebp/scikit-survival/blob/v0.20.0/tests/test_metrics.py) case but wasn't able to easily reproduce the issue running locally but in a more resource-constrained environment it is easily reproduced e.g. via this code snippet:

```
import numpy as np
from sksurv.metrics import concordance_index_censored

step=10000

for n_samples in range(10*step, 20*step, step):

    event_indicator = np.zeros(n_samples, dtype=bool)
    event_time = np.zeros(n_samples, dtype=int)
    for idx in range(n_samples):
        event_indicator[idx] = (idx % 2)
        event_time[idx] = 100 + idx//10
    estimate = np.zeros(n_samples, dtype=float)

    print(f"n_samples={n_samples} concordance_index_censored call before")
    concordance_index_censored(event_indicator, event_time, estimate)
    print(f"n_samples={n_samples} concordance_index_censored call after")
```

### DIFF
--- a/sksurv/metrics.py
+++ b/sksurv/metrics.py
@@ -103,11 +103,8 @@ def _get_comparable(event_indicator, event_time, order):
         censored_at_same_time = ~event_at_same_time
         for j in range(i, end):
             if event_indicator[order[j]]:
-                mask = np.zeros(n_samples, dtype=bool)
-                mask[end:] = True
-                # an event is comparable to censored samples at same time point
-                mask[i:end] = censored_at_same_time
-                comparable[j] = mask
+                mask_info = (i, end, censored_at_same_time)
+                comparable[j] = mask_info
                 tied_time += censored_at_same_time.sum()
         i = end
 
@@ -115,6 +112,7 @@ def _get_comparable(event_indicator, event_time, order):
 
 
 def _estimate_concordance_index(event_indicator, event_time, estimate, weights, tied_tol=1e-8):
+    n_samples = len(event_time)
     order = np.argsort(event_time)
 
     comparable, tied_time = _get_comparable(event_indicator, event_time, order)
@@ -128,7 +126,17 @@ def _estimate_concordance_index(event_indicator, event_time, estimate, weights, 
     tied_risk = 0
     numerator = 0.0
     denominator = 0.0
-    for ind, mask in comparable.items():
+    for ind, mask_info in comparable.items():
+        # mask info in three parts
+        i = mask_info[0]
+        end = mask_info[1]
+        censored_at_same_time = mask_info[2]
+        # construct (potentially large) mask from (smaller) mask info
+        mask = np.zeros(n_samples, dtype=bool)
+        mask[end:] = True
+        # an event is comparable to censored samples at same time point
+        mask[i:end] = censored_at_same_time
+
         est_i = estimate[order[ind]]
         event_i = event_indicator[order[ind]]
         w_i = weights[order[ind]]


sebp:
Indeed, the memory usage is quite excessive, great catch!

I believe you can optimize it even further by
1. converting `_get_comparable` to a generator via `yield`
2. Initializing an empty `mask` in `_estimate_concordance_index`, which you reset at the beginning of each loop `mask.fill(False)`.

cpoerschke:
Hi @sebp - thank you very much for reviewing and suggesting further optimisations!

generators: i had to [read up on that](https://docs.python.org/3/howto/functional.html#generators) and the changes i made are a first attempt then; tests continue to pass.

empty mask outside the loop and fill resetting inside the loop: i tried that but didn't observe a reduced number of allocations, perhaps a mask inside the loop can be efficiently 'recycled' if it's the same shape and dtype? allocation observation was via [memray's getting started](https://bloomberg.github.io/memray/getting_started.html#running-the-analysis) step and naive mouse-hover-for-tooltip manual inspection of the flame graph. having `np.create/mask.fill` instead of just `np.zeros` also marginally adds code reading complexity and so i've reverted that change again for now.

sksurv/metrics.py
```diff
@@ -103,32 +102,34 @@ def _get_comparable(event_indicator, event_time, order):
...
         for j in range(i, end):
             if event_indicator[order[j]]:
-                mask = np.zeros(n_samples, dtype=bool)
-                mask[end:] = True
-                # an event is comparable to censored samples at same time point
-                mask[i:end] = censored_at_same_time
-                comparable[j] = mask
+                mask_info = (i, end, censored_at_same_time)
                 tied_time += censored_at_same_time.sum()
+                yield (j, mask_info, tied_time)
```
cpoerschke, line 107: Actually ... with the `yield` here then it should be possible to return a mask, not mask info, let me go try that ...
```suggestion
                yield (j, mask, tied_time)
```

sebp:
> empty mask outside the loop and fill resetting inside the loop: i tried that but didn't observe a reduced number of allocations, perhaps a mask inside the loop can be efficiently 'recycled' if it's the same shape and dtype? allocation observation was via [memray's getting started](https://bloomberg.github.io/memray/getting_started.html#running-the-analysis) step and naive mouse-hover-for-tooltip manual inspection of the flame graph. having `np.create/mask.fill` instead of just `np.zeros` also marginally adds code reading complexity and so i've reverted that change again for now.

I don't know how numpy handles memory allocations either. Since you actually used a memory profiler and did not observe a difference, I would prefer the more readable code as well.

sksurv/metrics.py
```diff
@@ -89,7 +89,6 @@ def _check_estimate_2d(estimate, test_time, time_points, estimator):
 def _get_comparable(event_indicator, event_time, order):
```
sebp, line 89: Can you please rename `_get_comparable` to `_iter_comparable` to make it clearer what the function does.

Issue #133, opened 2020-07-20
How to plot the tree for a survival tree.


anirbanmishra:
I got the estimator using :
estimator = SurvivalTree().fit(X_train, y_train)

But when I use tree.plot_tree I get an error stating the following:

AttributeError: 'SurvivalTree' object has no attribute 'criterion'

I tried graphviz too and get the same error.

So basically I want the survival tree to be plotted for better interpretation.

sebp:
Duplicate of #110

Issue #110, opened 2020-04-25
viz of ensemble models


ad05bzag:
Hi!

would you have any advice on how to visualize decision path / decision trees from the ensemble survival model methods (either RF or Gradient Boosting)?

sebp:
I uploaded a modification of the sklearn.tree.export module to [this gist](https://gist.github.com/sebp/82f54a5480e170fd3c303873b75a07ff). It allows you to use the `plot_tree` function with an instance of `SurvivalTree`.

```python
plot_tree(survival_tree, feature_names=feature_names, impurity=False, label="none")
```

ad05bzag:
cool! thank you so much. I ended up making it work using three methods: your suggested one , export_graphviz (graphviz library), dtreeviz (explained.ai package). 

Your worked like so:
`
rsf = RandomSurvivalForest()
plot_tree(rsf.estimators_[0], feature_names=X_train.iloc[:,2:].columns.tolist(), impurity=False, label="none")
`
For two latter methods, it came down to just subsetting the estimators, so like 

* graphviz:

```
from graphviz import Source
from sklearn.tree import export_graphviz

gbs=GradientBoostingSurvivalAnalysis()
feature_names=X_train.iloc[:,2:].columns.tolist()
target_names=df['death'].unique()

graph=Source(export_graphviz(gbs.estimators_[0][0],
                             feature_names=feature_names,
                             class_names=target_names,
                             filled=True, 
                             rounded=True))
```


* dtreeviz
```
from dtreeviz.trees import *

viz = dtreeviz(gbs.estimators_[0][0], X_trainRF, np.array([x[1] for x in y_trainRF]), target_name='time to event',
               feature_names=X_train.iloc[:,2:].columns.tolist())
```


Q: what I have been thinking about is that the innate limitation of all these methods is that we are checking only a single tree in all three methods. Which seems lacking? Both for gradient boosting and random forests i have >100 estimators. How to give a more cohesive, inclusive picture?

sebp:
I'm not aware of a method that visualizes ensembles of trees in a easy to understand way.

julienbeisel:
@sebp Thanks for this Gist, it's very useful !

However, i'm having troubles showing impurity, every value is "inf". Do you know if it is possible to display the logrank value for each node? Thank you.

sebp:
[...]

PR #366, opened 2023-06-02
set self.criterion in SurvivalTree.fit

cpoerschke:
**Checklist**

- [X] pytest passes
- [ ] tests are included
- [X] code is well formatted

An `AttributeError: 'SurvivalTree' object has no attribute 'criterion'` error is encountered as mentioned in #110 and #133 issues. Setting the `criterion` in the `fit` method would allow plotting of the tree, with the caveat that the tree structure is visible but the `impurity` and `value` values are `inf` and arrays.

Code snippet:

```
import numpy as np
import sklearn.tree
import sksurv.tree
import sksurv.util

X = np.array([[80, 20], [70, 30], [60, 40], [50, 50], [40, 60], [30, 70], [20, 80]])
y = sksurv.util.Surv.from_arrays(time=(1, 2, 3, 4, 5, 6, 7),
                                 event=(True, True, True, False, False, False, False))

st = sksurv.tree.SurvivalTree()
st.fit(X, y)

sklearn.tree.plot_tree(st, feature_names=["one", "two"])
```

### DIFF
--- a/sksurv/tree/tree.py
+++ b/sksurv/tree/tree.py
@@ -229,14 +229,15 @@ class SurvivalTree(BaseEstimator, SurvivalAnalysisMixin):
         self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * 2
 
         # Build tree
-        self.criterion = LogrankCriterion(self.n_outputs_, n_samples, self.event_times_)
+        self.criterion = "logrank"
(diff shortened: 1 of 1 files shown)


sebp:
Thanks for raising this issue.

Rather then setting `self.criterion` to an instance of `LogrankCriterion`, you can set it to the string `"logrank"`. This way the type matches that of sklearn's tree estimators.

PR #369, opened 2023-06-08
reduce SurvivalTree.predict's memory use

cpoerschke:
**Checklist**

- [X] pytest passes
- [X] tests are included
- [X] code is well formatted

issue encountered: Passing relatively lots of technical data samples to `SurvivalTree.predict` results in "kernel died" in a notebook environment.

## analysis

The `pred = self.tree_.predict(X)` call in `SurvivalTree.predict_cumulative_hazard_function` returns an array of `(n_samples, self.event_times_, 2)` shape which is then returned as a `(n_samples, self.event_times_)` shape and then `SurvivalTree.predict` does a `.sum(1)` on it, resulting in an overall `(n_samples)` shape return value. The `n_samples X self.event_times_` shapes use a lot of memory for larger samples especially if the number of event times is a proportion of samples.

## proposed solution

* Do the `self.tree_.predict` call on a one-row-at-a-time basis, requiring one `(1, self.event_times_, 2)` shape for each iteration.
* Have `SurvivalTree.predict_cumulative_hazard_function` do the `sum`-ing for each row, removing the need for a `(n_samples, self.event_times_)` shape array.

## testing

peak memory usage stats inspected before and after the change using `memray run ./pr6-demo.py` profiling.

code snippet:

```
#!/usr/bin/env python

import numpy as np
import sksurv.tree
import sksurv.util

n = 12345

verbose = (n <= 10)

feature1 = np.arange(0, n, 1)
feature2 = n - feature1

times = np.arange(0,n) + 1
events = (times < times[len(times)//2])

X = np.vstack((feature1, feature2)).T
y = sksurv.util.Surv.from_arrays(time=times, event=events)

if verbose:
    print(f"X={X}\ny={y}")
else:
    print(f"X.shape={X.shape}\ny.shape={y.shape}")

st = sksurv.tree.SurvivalTree(max_leaf_nodes=100)
print(st.fit(X, y))

risk_scores = st.predict(X)

if verbose:
    print(f"risk_scores={risk_scores}")
else:
    print(f"risk_scores.shape={risk_scores.shape}")
```


sksurv/tree/tree.py
```diff
@@ -369,8 +369,7 @@ def predict(self, X, check_input=True):
         risk_scores : ndarray, shape = (n_samples,)
             Predicted risk scores.
         """
-        chf = self.predict_cumulative_hazard_function(X, check_input, return_array=True)
-        return chf.sum(1)
+        return self.predict_cumulative_hazard_function(X, check_input, return_array="sum")
```
cpoerschke, line 372: This changes all existing behaviour; an alternative would be to have an opt-in constructor flag so that callers need to opt-in to get the new behaviour.

sksurv/tree/tree.py
```diff
@@ -432,6 +434,13 @@ def predict_cumulative_hazard_function(self, X, check_input=True, return_array=F
         check_is_fitted(self, 'tree_')
         X = self._validate_X_predict(X, check_input, accept_sparse="csr")
 
+        if return_array == "sum":
+            n_samples = X.shape[0]
+            array_of_sums = np.empty(n_samples)
+            for idx in range(0, n_samples):
+                array_of_sums[idx] = self.tree_.predict(X[idx:idx+1])[..., 0].sum()
```
cpoerschke, line 441: This iterates on a one-row-at-a-time basis; an alternative could be to iterate over multiple-rows-at-a-time, somehow, I haven't yet explored how that might work in numpy or so.

sebp:
Thanks for your PR.

I agree that `predict` is quite heavy on memory usage. Part of it has to do with sklearn's Tree-related code, where predictions are returned via the split criterion's [node_value](https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/tree/_criterion.pxd#L62-L65) method. It is currently implemented to return the full survival function and cumulative hazard function for each sample, disregarding whether `predict`, `predict_survival_function`, or `predict_cumulative_hazard_function` has been called. I could imagine adding a low-memory option that disables computing survival and cumulative hazard function (CHF) and just returns the event counts (sum over CHF, i.e. what predict returns).

cpoerschke:
> ... Part of it has to do with ... I could imagine adding a low-memory option that disables computing survival and cumulative hazard function (CHF) and just returns the event counts (sum over CHF, i.e. what predict returns).

Thanks for the context and quick feedback!

I've added a `low_memory=False` option in the latest commit, though I guess its current use does not disable the computations as such. And of course for any new option there should be test coverage too.

cpoerschke:
> ... Part of it has to do with sklearn's Tree-related code, where predictions are returned via the split criterion's [node_value](https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/tree/_criterion.pxd#L62-L65) method. It is currently implemented to return the full survival function and cumulative hazard function for each sample, disregarding whether `predict`, `predict_survival_function`, or `predict_cumulative_hazard_function` has been called. I could imagine adding a low-memory option that disables computing survival and cumulative hazard function (CHF) and just returns the event counts (sum over CHF, i.e. what predict returns).

So I haven't worked with .pxd and .pyx code before but from a little bit of code reading just now ... is the idea conceptually that in low-memory mode:
* https://github.com/sebp/scikit-survival/blob/v0.20.0/sksurv/tree/tree.py#L227-L229 would request fewer outputs and classes from `LogRankCriterion` and `Tree`
* https://github.com/sebp/scikit-survival/blob/v0.20.0/sksurv/tree/_criterion.pyx#L305-L327 would fill the reduced size `dest` with the event counts
* `predict_cumulative_hazard_function` and `predict_survival_function` cannot be used in low-memory mode
* some but not all of the above and/or something else?

sksurv/tree/tree.py
```diff
@@ -225,8 +231,9 @@ def fit(self, X, y, sample_weight=None, check_input=True):
         params = self._check_params(n_samples)
 
         self.n_outputs_ = self.event_times_.shape[0]
-        # one "class" for CHF, one for survival function
-        self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * 2
+        # one "class" for CHF, one for survival function, one as placeholder
+        j_delta = 3
+        self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * j_delta
```
cpoerschke, line 236: This is an interim development step change i.e. my thinking is along the following lines:
* step 1: change computation code to compute a third value that matches the first value (but don't use the third value as yet)
* step 2: change using code to use the third value instead of the first value (and still do sum-ing)
* step 3: change code to output differently shaped values (in low memory mode)

sksurv/tree/_criterion.pyx
```diff
@@ -315,13 +315,14 @@ cdef class LogrankCriterion(Criterion):
         dest[0] = ratio  # Nelson-Aalen estimator
         dest[1] = 1.0 - ratio  # Kaplan-Meier estimator
 
-        j = 2
+        j_delta = 3
```
cpoerschke, line 318: Changing from delta=2 to delta=3 only here seems to be insufficient i.e. probably some other changes somewhere are needed too but I haven't yet figured out what/where.

cpoerschke (commented):
Added work-in-progress notes inline.

sksurv/tree/_criterion.pyx
```diff
@@ -315,13 +315,14 @@ cdef class LogrankCriterion(Criterion):
...
+        j = j_delta
         for k in range(1, self.n_event_times):
             rs = &self.riskset_total[k]
-            dest[j] = dest[j - 2]
-            dest[j + 1] = dest[j - 1]
+            dest[j] = dest[j - j_delta]
```
sebp, line 322: `dest` is a pointer to the array that hold the predictions. It's a 2d array of size `n_unique_times × 2`. The first column is the survival function, the second column the cumulative hazard function. This is also why incrementing by 2 is required.

In low-memory mode, `dest` should have `shape = (2,)` by changing `SurvivalTree.n_classes_` and `SurvivalTree.n_outputs_` accordingly. Then, `dest[0]` should hold the sum over the CHF, and `dest[1]` the sum over the survival function.

Note that the code is going to change with #371. Once it is merged, CHF and survival function will be over all unique time points (not just event times). Thus, the summing operation also needs to consider at which times an event was observed.

cpoerschke: Thanks @sebp for the pointers here!

I've made changes and it almost works now, only the _"the summing operation also needs to consider at which times an event was observed"_ bit remains to be figured out.

sksurv/tree/tree.py
```diff
@@ -229,6 +236,11 @@ def fit(self, X, y, sample_weight=None, check_input=True):
         # one "class" for CHF, one for survival function
         self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * 2
 
+        if self.low_memory:
```
cpoerschke, line 239: an alternative here would be
```
if self.low_memory:
    ...
else:
    ...
```
and likewise elsewhere

sksurv/tree/_criterion.pyx
```diff
@@ -328,12 +328,29 @@ cdef class LogrankCriterion(Criterion):
...
+            dest_j1 = dest[1]
+            for i in range(1, self.n_unique_times):
+                self.riskset_total.at(i, &n_at_risk, &n_events)
+                if n_at_risk != 0:
+                    ratio = n_events / n_at_risk
+                    dest_j0 += ratio
+                    dest_j1 *= 1.0 - ratio
+                if True: # TODO: only sum for event times
+                    dest[0] += dest_j0
+                    dest[1] += dest_j1
```
sebp, line 351: The sum over the survival function can be skipped too, the sum over the CHF (`dest_j0`) is sufficient.

If the `i`-th unique time does not have an event, you would not add anything to the sum. The corresponding Python code is https://github.com/sebp/scikit-survival/blob/bd2e24029632c4a0dbccff86ee905ba8d9b6f5e7/sksurv/tree/tree.py#L368
where `is_event_time` is created at https://github.com/sebp/scikit-survival/blob/bd2e24029632c4a0dbccff86ee905ba8d9b6f5e7/sksurv/tree/tree.py#L215
It is currently not accessible from Cython, so you would need to pass it to `LogrankCriterion`.

sksurv/tree/tree.py
```diff
@@ -229,6 +236,11 @@ def fit(self, X, y, sample_weight=None, check_input=True):
...
+            self.n_outputs_ = 1
+            # one "class" for the sum over the CHF, one for the sum over the survival function
+            self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * 2
```
sebp, line 242: Since you only need a single output (the sum over the CHF), it should be sufficient to set `n_outputs_` to 1 and don't set `self.n_classes_` at all.

sksurv/tree/tree.py
sebp: I would prefer if-else

sksurv/tree/tree.py
```diff
@@ -491,6 +514,10 @@ def predict_survival_function(self, X, check_input=True, return_array=False):
         >>> plt.ylim(0, 1)
         >>> plt.show()
         """
+
+        if self.low_memory:
+            raise NotImplementedError("predict_survival_function is not implemented in low memory mode.")
```
sebp, line 519: Can be combined with the check in `predict_cumulative_hazard_function` by adding a separate method `_check_low_memory`.

tests/test_tree.py
```diff
@@ -779,3 +779,36 @@ def test_predict_sparse(make_whas500):
...
+    tree0 = SurvivalTree(min_samples_leaf=10, random_state=seed, low_memory=False)
+    tree0.fit(X_train, y_train)
+    y_pred_0 = tree0.predict(X_test)
+
+    tree1 = SurvivalTree(min_samples_leaf=10, random_state=seed, low_memory=True)
+    tree1.fit(X_train, y_train)
+    y_pred_1 = tree1.predict(X_test)
+
+    assert y_pred_0.shape[0] == X_test.shape[0]
+    assert y_pred_1.shape[0] == X_test.shape[0]
```
sebp, line 804: `assert_array_almost_equal` checks shapes too

tests/test_tree.py
```diff
@@ -779,3 +779,36 @@ def test_predict_sparse(make_whas500):
...
     assert_array_equal(y_surv, y_surv_csr)
+
+
+def test_predict_low_memory(make_whas500):
+    seed = 42
+    whas500 = make_whas500(to_numeric=True)
+    X, y = whas500.x, whas500.y
+    # Duplicates values in whas500 leads to assert errors because of
+    # tie resolution during tree fitting.
+    # Using a synthetic dataset resolves this issue.
```
sebp, line 790: Actually, ties should not cause any problems.

sksurv/tree/tree.py
```diff
@@ -424,6 +443,10 @@ def predict_cumulative_hazard_function(self, X, check_input=True, return_array=F
         >>> plt.ylim(0, 1)
         >>> plt.show()
         """
+
+        if self.low_memory:
+            raise NotImplementedError("predict_cumulative_hazard_function is not implemented in low memory mode.")
```
sebp, line 448: Please add "run fit with low_memory=False to disable low memory mode."

sebp (changes requested):

sksurv/tree/tree.py
```diff
@@ -225,13 +232,19 @@ def fit(self, X, y, sample_weight=None, check_input=True):
...
+        else:
+            self.n_outputs_ = self.unique_times_.shape[0]
+            # one "class" for CHF, one for survival function
+            self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp) * 2
 
         # Build tree
         self.criterion = "logrank"
-        criterion = LogrankCriterion(self.n_outputs_, n_samples, self.unique_times_)
+        criterion = LogrankCriterion(self.n_outputs_, n_samples, self.unique_times_,
+                                     self.is_event_time_.astype(np.int32))
```
cpoerschke, line 247:
```suggestion
                                     self.is_event_time_)
```

Tried this ---^ but it gives this error:

```
>   def __cinit__(self, SIZE_t n_outputs, SIZE_t n_samples, const DOUBLE_t[::1] unique_times, const bint[::1] is_event_time):
E   ValueError: Buffer dtype mismatch, expected 'const bool' but got 'bool'
```

sksurv/tree/tree.py
cpoerschke: Just noting that not setting `self.n_classes_` or setting it to `1` gave an error, hence only removed the ` * 2` to give `self.n_classes_ = np.ones(self.n_outputs_, dtype=np.intp)` now then for a single output.

tests/test_tree.py
cpoerschke: Just noting that this shape check is against `X_test` (like in the `test_predict_sparse` test) since the `assert_array_almost_equal` check would only check one `y_pred_*` against the other `y_pred_*` shape. I guess technically one of the two checks against `X_test` shape could be omitted but it's probably clearer to just check both.

cpoerschke (commented):
Thanks @sebp for the feedback! I think I've addressed all points and tests pass locally. Please let me know if there's any additional suggestions. Thank you.

sksurv/tree/_criterion.pyx
```diff
@@ -328,19 +330,31 @@ cdef class LogrankCriterion(Criterion):
...
+
+        # low memory mode
+        if  self.n_outputs == 1:
+            dest[0] = dest_j0 = 0
+            for i in range(0, self.n_unique_times):
+                self.riskset_total.at(i, &n_at_risk, &n_events)
+                if n_at_risk != 0:
+                    ratio = n_events / n_at_risk
+                    dest_j0 += ratio
+                if self.is_event_time[i]:
```
sebp, line 343: You should assign `self.is_event_time` a local variable first, because otherwise each call will have some overhead. You can see that if you run `cython -a _criterion.pyx` and have a look the generated HTML file. If a line is highlighted yellow, there is some overhead.

sebp (approved):

PR #385, opened 2023-06-27
support predict_(cumulative_hazard|survival)_function in Stacking

### LINKED ISSUE #368: Add support for predict_survival_function to Stacking
As mentioned in #364, `Stacking` currently does not support `predict_survival_function` nor `predict_cumulative_hazard_function`.

If the meta-estimator supports these functions, so should `Stacking`.

cpoerschke:
**Checklist**

- [X] closes #368 
- [X] pytest passes
- [X] tests are included
- [X] code is well formatted

For the #368 issue.

### DIFF
--- a/sksurv/meta/ensemble_selection.py
+++ b/sksurv/meta/ensemble_selection.py
@@ -63,9 +63,6 @@ class MeanEstimator(BaseEstimator):
     def predict(self, X):  # pylint: disable=no-self-use
         return X.mean(axis=X.ndim - 1)
 
-    def predict_survival_function(self, X):
-        return X.mean(axis=X.ndim - 1)
-
 
 class MeanRankEstimator(BaseEstimator):
     def fit(self, X, y=None, **kwargs):  # pragma: no cover; # pylint: disable=unused-argument
--- a/sksurv/meta/ensemble_selection.py
+++ b/sksurv/meta/ensemble_selection.py
@@ -63,6 +63,9 @@ class MeanEstimator(BaseEstimator):
     def predict(self, X):  # pylint: disable=no-self-use
         return X.mean(axis=X.ndim - 1)
 
+    def predict_survival_function(self, X):
+        return X.mean(axis=X.ndim - 1)
+
 
 class MeanRankEstimator(BaseEstimator):
     def fit(self, X, y=None, **kwargs):  # pragma: no cover; # pylint: disable=unused-argument
--- a/sksurv/meta/stacking.py
+++ b/sksurv/meta/stacking.py
@@ -289,12 +289,6 @@ class Stacking(MetaEstimatorMixin, SurvivalAnalysisMixin, _BaseComposition):
 
     @property_available_if(_meta_estimator_has("unique_times_"))
     def unique_times_(self):
-        """
-        Returns
-        -------
-        unique_times_ : array of shape = (n_unique_times,)
-            Unique time points.
-        """
         return self.meta_estimator.unique_times_
 
     @available_if(_meta_estimator_has("predict_cumulative_hazard_function"))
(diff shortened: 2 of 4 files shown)


sksurv/meta/ensemble_selection.py
```diff
@@ -63,6 +63,9 @@ def fit(self, X, y=None, **kwargs):  # pragma: no cover; # pylint: disable=unuse
     def predict(self, X):  # pylint: disable=no-self-use
         return X.mean(axis=X.ndim - 1)
 
+    def predict_survival_function(self, X):
+        return X.mean(axis=X.ndim - 1)
```
sebp, line 67: Taking the average over survival functions is problematic, because you could end up with an average survival function that is not monotonically decreasing.

sebp (changes requested):
Thanks for taking this on!

In this setting, the meta estimator needs to something more clever than a simple average. It should be any existing estimator that implements `predict_survival_function`, like `CoxPHSurvivalAnalysis`. In this setting, the predicted risk scores of the base models become the features of the meta estimator.

cpoerschke (commented):
Thanks @sebp for the quick feedback!

The latest commit reverts the `MeanEstimator` change and amends the test to match.

`return_array=False` also added to `Stacking.predict_survival_function` now then.

sksurv/meta/stacking.py
```diff
@@ -285,3 +285,31 @@ def predict_log_proba(self, X):
...
+        -------
+        survival : ndarray
+            If `return_array` is set, an array with the probability of
+            survival for each `self.unique_times_`, otherwise an array of
+            length `n_samples` of :class:`sksurv.functions.StepFunction`
+            instances will be returned.
+
+        return_array : boolean, default: False
+            If set, return an array with the probability
+            of survival for each `self.unique_times_`,
```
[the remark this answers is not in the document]
sebp: Could you forward unique_times_ too (if it exists)? Note that it is an attribute, which will work differently, see https://github.com/sebp/scikit-survival/blob/65ffdf9f4262527f4cadc91f0b011c0fdecb77f4/sksurv/util.py#L303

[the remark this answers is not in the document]
cpoerschke: Thanks for the pointer! https://github.com/sebp/scikit-survival/pull/385/commits/fa6c11e3413558e05e2988631c2354b45a927190 adds forwarding with matching test.

tests/test_stacking.py
```diff
@@ -234,6 +238,78 @@ def test_predict_proba(method):
...
+            CoxPHSurvivalAnalysis(),
+            [("rsf", RandomSurvivalForest(random_state=0)), ("svm", FastSurvivalSVM(random_state=0))],
+            probabilities=False,
+        )
+        with pytest.raises(AttributeError, match="'BreslowEstimator' object has no attribute 'unique_times_'"):
+            meta.unique_times_  # pylint: disable=pointless-statement
+
+        whas500 = make_whas500(with_mean=False, with_std=False, to_numeric=True)
+        meta.fit(whas500.x_data_frame, whas500.y)
+        assert 395 == len(meta.unique_times_)  # pylint: disable=pointless-statement
```
cpoerschke, line 261:
```suggestion
        assert 395 == len(meta.unique_times_)
```

sksurv/meta/stacking.py
```diff
@@ -285,3 +286,61 @@ def predict_log_proba(self, X):
         """
         Xt = self._predict_estimators(X)
         return self.final_estimator_.predict_log_proba(Xt)
+
+    @property_available_if(_meta_estimator_has("unique_times_"))
+    def unique_times_(self):
```
sebp, line 291: Could you please add a API doc string for `unique_times_`too?

cpoerschke: Hmm, adding docs somehow doesn't work i.e. it seems the doc generation isn't "aware" of the conditionality of the attribute being documented. So I've for now undone the changes. Interestingly for the conditional methods there doesn't seem to be the same issue it seems.

sebp:
There's currently an issue with the sphinx plugin `sphinx.ext.linkcode`, which fails when using the `property_available_if` decorator.

sebp:
@cpoerschke Thanks for your contributions. Your changes have been merged via #388

PR #430, opened 2024-01-25
Corrected kaplan_meier_estimator documentation to give correct default argument

CaderIdris:
Problem: kaplan_meier_estimator docstring shows conf_type default argument as log-log instead of None
Solution: Changed docstring to match code

**Checklist**

- [x] code is well formatted

Was wondering why my code  wasn't working, turns out the wrong default argument is listed on the docs. This should fix it. Haven't checked for disparity between docs and code for other functions/classes but this is the only one I've had issues with so far.


sebp:
Thanks for your PR. Could you please also fix the documentation of `sksurv.nonparametric.SurvivalFunctionEstimator`?

Issue #448, opened 2024-04-06
[BUG] `GradientBoostingSurvivalAnalysis` - docstring/logic mismatch on possible `criterion` values


fkiraly:
In `GradientBoostingSurvivalAnalysis`, docstring and logic do not agree on the set of possible `criterion` values.

Docstring implies that `"friedman_mse"`, `"mse"`, and `"mae"` are valid, but passing `"mse"` or `"mae"` triggers a `ValueError` which says that only `"friedman_mse"` and `"squared_error"` are valid.

Indeed, the constraints from `BaseGradientBoosting` have `"criterion": [StrOptions({"friedman_mse", "squared_error"})],`.

This seems consistent with the construction of `DecisionTreeRegressor` under the hood (which would allow additional options, though).

sebp:
You are right that `sklearn.tree.DecisionTreeRegressor` does support "absolute_error" and "poisson", however the gradient boosting models in sklearn do not, e.g. `sklearn.ensemble.GradientBoostingClassifier`. Not sure why, but I would stick to the same options that gradient boosting models in sklearn support.

PR #449, opened 2024-04-06
[DOC] fix `GradientBoostingSurvivalAnalysis` docstring on `criterion` parameter

fkiraly:
Fixes https://github.com/sebp/scikit-survival/issues/448, under the assumption that the issue is with the docstring, not the logic.

### DIFF
--- a/sksurv/ensemble/boosting.py
+++ b/sksurv/ensemble/boosting.py
@@ -682,11 +682,11 @@ class GradientBoostingSurvivalAnalysis(BaseGradientBoosting, SurvivalAnalysisMix
         and an increase in bias.
         Values must be in the range `(0.0, 1.0]`.
 
-    criterion : string, optional, default: 'friedman_mse'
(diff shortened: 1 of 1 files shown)


sebp (changes requested):
Thanks for noticing!

For consistence with sklearn, and to make it easier to track changes, I'd prefer to copy the API doc from sklearn. This would also remove "mae", which isn't supported by `BaseGradientBoosting`.

https://github.com/scikit-learn/scikit-learn/blob/f07e0138bfee41cd2c0a5d0251dc3fe03e6e1084/sklearn/ensemble/_gb.py#L1159-L1164

fkiraly:
of course, done

fkiraly:
> Thanks for noticing!

PS, it was actually the tests in `skpro` that noticed.

sebp:
@fkiraly Thank you. Merged in commit f5e7c04

fkiraly:
... why did you close this and merge a copy??

sebp:
I squashed your commits before merging.

PR #484, opened 2024-10-21
Disable early stopping criteria

### LINKED ISSUE #41: Alpha Behavior
When I run a model like this example:
__mod = coxnet.CoxnetSurvivalAnalysis(n_alphas=30, l1_ratio=1.0)__
There are times (and it could be data specific) where the length of the coefficients out of the model are less than the n_alphas I specified. For example, it often stops at 5 alphas deep. 
The paths might have 15 variables > 0 (coming out) at 5 alphas deep, which is fine. The strange thing I am seeing is, lets say I set n_alphas=20 on the same data set. I end up getting __more__ variables > 0 along the path (and still stopping at 5 alphas deep) or vice verse, if I set n_alphas = 40 on the same data set, I end up getting __less__ variables > 0 along the path and once again the algorithm is automatically stopping at 5 alphas deep.  (Im referring to the parameters as variables.)

Im assuming this is a bug as the way I have working with Elastic Nets in the past is that the alpha sequence should exponentially decrease toward some min but I should see more variables > 0 as I move forward and closer to the min in this alpha curve. Such that if I see 15 variables at alpha=30, then I should see <15 at alpha<30 and the reverse.

Could there be some ratio somewhere that is picking up a variable of similar name to a global that is confusing the alpha parameters in the Elastic net?

Sann5:
**Checklist**

- [x] closes #41 
- [x] <del>pytest passes</del> pytest (that were passing before) passes
- [x] tests are included (no tests added)
- [x] code is well formatted
- [ ] documentation renders correctly (this idk)

In `sksurv/linear_model/src/coxnet/coxnet.h` we find the `fit` method for the `CoxnetSurvivalAnalysis` class. This fits a model for the desired regularization path. The current implementation includes an early stop condition which stops fiting a model for the given alphas based on a deviance ratio. However it is desirable sometimes to fit the whole path (using all the alphas) even if the performance does not change significanly across alphas (see #41 for a more detailed explanation of the behaviour). The obvios solution is to get rid of the early stop condition (which is what I have done here by commentitg it out), however I think the better solution would be to include a boolean flag in the class definition that specifies if the user wishes or not to use this early stopping functionality. Sadly I do not have the time to implement such a solution. Please feel free to use this PR as a starting point.

### DIFF
--- a/sksurv/linear_model/src/coxnet/coxnet.h
+++ b/sksurv/linear_model/src/coxnet/coxnet.h
@@ -249,13 +249,13 @@ void Coxnet<T, S, U>::fit(
         coef_path.col(i) = m_params.coef_x;
         dev_ratio_path[i] = 1. - (loglik_saturated - log_likelihood()) / deviance_null;
 
-        if (i >= Constants<Scalar>::MIN_ALPHAS()) {
-            /* abort if change in deviance ratio was small, compared to previous (larger) alpha */
-            if (dev_ratio_path[i - Constants<Scalar>::MIN_ALPHAS() + 1] / dev_ratio_path[i] > 0.999) {
-                ++i;
-                break;
-            }
-        }
+        // if (i >= Constants<Scalar>::MIN_ALPHAS()) {
+        //     /* abort if change in deviance ratio was small, compared to previous (larger) alpha */
+        //     if (dev_ratio_path[i - Constants<Scalar>::MIN_ALPHAS() + 1] / dev_ratio_path[i] > 0.999) {
+        //         ++i;
+        //         break;
+        //     }
+        // }
     }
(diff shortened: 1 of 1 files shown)


sebp:
You are correct that this part of the code is probably the reason why coefficients for at most 5 alphas are returned.

However, this check is also part of the reference implementation in the R glmnet package: https://github.com/cran/glmnet/blob/ace3461223a8fe8a3e0c2f749d3d7e8e84e49ea8/src/glmnet3.f90#L2489

Hence, the code does have its purpose. The main question where the two implementations deviate from each other. I would suggest to use an example where sksurv terminates early and compare the steps in glmnet and sksurv with a debugger.

PR #520, opened 2025-04-03
chore(metrics): fixing and ending bracket typo in ValueError for check_times

sentisso:
**Checklist**

- [x] pytest passes
- [x] code is well formatted

Fixing a minor typo in a `ValueError` for the `_check_times` validation function. Previous implementation yielded error:
```
ValueError: all times must be within follow-up time of test data: [3.0; 2880.0[
```

Corrected version:
```
ValueError: all times must be within follow-up time of test data: [3.0; 2880.0]
```

### DIFF
--- a/sksurv/metrics.py
+++ b/sksurv/metrics.py
@@ -65,7 +65,7 @@ def _check_times(test_time, times):
 
     if times.max() >= test_time.max() or times.min() < test_time.min():
         raise ValueError(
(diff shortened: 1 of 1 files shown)


sebp:
Thanks for helping to improve scikit-survival.

I believe the current error message ist correct, because the upper bound of the interval is excluded, which is indicated by `[`. `]` would indicate that it is included.

PR #529, opened 2025-06-11
Update understanding_predictions.rst

dryezl:
spell corrected


sebp:
Thank you for helping to improve scikit-survival.

I've been using the European notation for half-open intervals throughout the documentation. Hence, I prefer to keep the notation `[a, b[` for consistency.

PR #590, opened 2026-05-21
fix: treat pandas Boolean as numeric in clinical_kernel

### LINKED ISSUE #589: Bug:  clinical_kernel silently drops pandas Boolean columns; ClinicalKernelTransform.fit raises TypeError on the same input
### Describe the bug

The two pandas-input entry points of `sksurv.kernels` disagree on how to handle Boolean columns, and **both** are broken:

- `clinical_kernel(df)` silently drops Boolean columns from its feature set, but still divides the kernel matrix by `x.shape[1]`. The result is a kernel matrix biased by `(n_features - n_bool_cols) / n_features` with no warning. For a `(float64, bool)` frame the diagonal collapses from `1.0` to `0.5`.
- `ClinicalKernelTransform().fit(df)` on the same frame raises `TypeError: numpy boolean subtract, the '-' operator, is not supported`.

Root cause:

- `_get_continuous_and_ordinal_array` uses `x.select_dtypes(include=[np.number])` to find continuous columns. In numpy's type hierarchy `np.bool_` is a sibling of `np.number`, not a subclass, so Boolean columns match neither the numeric filter nor the `object`/`category` filter and are dropped.
- `_prepare_by_column_dtype` uses `pandas.api.types.is_numeric_dtype`, which returns `True` for Boolean. The column enters the numeric branch but `col.max() - col.min()` then fails because `numpy` removed Boolean subtraction.

The polars-aware paths in this module already treat `nw.Boolean` as numeric, and pandas' own `is_numeric_dtype(bool)` returns `True`, so the established policy is "Boolean is numeric". The pandas paths are the lone outliers.

This is structural and does not depend on the pandas version. The crash in `ClinicalKernelTransform.fit` requires numpy >= 1.25 (older numpy emitted a `DeprecationWarning` instead), but the `clinical_kernel` silent-drop behavior is present regardless of numpy/pandas version.

Reproduced from a clean checkout of `main` at `6a1df8a` (= upstream `sebp/scikit-survival` main HEAD at time of filing).

I have a fix and tests ready and will open a PR linked to this issue.

### Steps/Code to Reproduce

```python
import numpy as np
import pandas as pd
from sksurv.kernels import clinical_kernel, ClinicalKernelTransform

df = pd.DataFrame({
    "age":   [20.0, 23.0, 26.0, 54.0, 100.0],
    "event": [True, False, True, True, False],   # bool dtype
})
df_uint = df.assign(event=df["event"].astype(np.uint8))  # same data, uint8 dtype

print(clinical_kernel(df))
print(clinical_kernel(df_uint))
print(np.allclose(clinical_kernel(df), clinical_kernel(df_uint)))

ClinicalKernelTransform().fit(df)
```

### Actual Results

```
>>> clinical_kernel(df)  # bool dtype
[[0.5     0.48125 0.4625  0.2875  0.     ]
 [0.48125 0.5     0.48125 0.30625 0.01875]
 [0.4625  0.48125 0.5     0.325   0.0375 ]
 [0.2875  0.30625 0.325   0.5     0.2125 ]
 [0.      0.01875 0.0375  0.2125  0.5    ]]

>>> clinical_kernel(df_uint)  # uint8 dtype
[[1.      0.48125 0.9625  0.7875  0.     ]
 [0.48125 1.      0.48125 0.30625 0.51875]
 [0.9625  0.48125 1.      0.825   0.0375 ]
 [0.7875  0.30625 0.825   1.      0.2125 ]
 [0.      0.51875 0.0375  0.2125  1.     ]]

>>> np.allclose(clinical_kernel(df), clinical_kernel(df_uint))
False

>>> ClinicalKernelTransform().fit(df)
Traceback (most recent call last):
  File "<stdin>", line 23, in <module>
  File ".../sksurv/kernels/clinical.py", line 256, in fit
    self._prepare_by_column_dtype(X)
  File ".../sksurv/kernels/clinical.py", line 208, in _prepare_by_column_dtype
    numeric_ranges.append(col.max() - col.min())
                          ~~~~~~~~~~^~~~~~~~~~~
TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead.
```

Note that with `bool` the diagonal of the kernel matrix is `0.5` instead of `1.0`, because the Boolean column is silently dropped but `x.shape[1] == 2` is still used for normalization. Downstream models (e.g. `FastKernelSurvivalSVM`) train and predict without any indication that the kernel is biased.

### Expected Results

`clinical_kernel(df)` and `clinical_kernel(df_uint)` should return the same matrix (Boolean and `uint8` are semantically equivalent here), and `ClinicalKernelTransform().fit(df)` should succeed and produce the same `X_fit_` / `_numeric_ranges` as the `uint8` variant. In other words, Boolean columns should be treated as numeric, consistent with `pandas.api.types.is_numeric_dtype(bool) is True` and with the polars path.

### Installed Versions

```shell
SYSTEM
------
Platform          : macOS-15.7.3-arm64-arm-64bit
Python version    : CPython 3.12.12
Python interpreter: .../.venv/bin/python3

DEPENDENCIES
------------
scikit-survival   : 0.27.1.dev18+g6a1df8a66
scikit-learn      : 1.8.0
numpy             : 2.4.6
scipy             : 1.17.1
pandas            : 3.0.3
numexpr           : 2.14.1
ecos              : 2.0.14
osqp              : 1.1.1
joblib            : 1.5.3
matplotlib        : None
pytest            : 9.0.3
sphinx            : None
Cython            : 3.2.4
pip               : None
setuptools        : 82.0.1
```

55Kamiryo:
**Checklist**

- [x] closes #589
- [x] pytest passes
- [x] tests are included
- [x] code is well formatted

Fixes the two divergent Boolean-handling bugs in `sksurv/kernels/clinical.py` described in #589:

1. **`clinical_kernel` silent drop.** `_get_continuous_and_ordinal_array` used `x.select_dtypes(include=[np.number])`, which excludes `np.bool_` because numpy treats it as a sibling of `np.number` rather than a subclass. Boolean columns matched neither the numeric nor the `object`/`category` filter, were silently dropped, and the resulting matrix was biased by `(n_features - n_bool_cols) / n_features` because normalization still used `x.shape[1]`.

   *Fix:* `select_dtypes(include=[np.number, "bool"])`.

2. **`ClinicalKernelTransform.fit` TypeError.** `_prepare_by_column_dtype` uses `pandas.api.types.is_numeric_dtype`, which returns `True` for Boolean, but `col.max() - col.min()` then failed with `TypeError: numpy boolean subtract` on numpy ≥ 1.25.

   *Fix:* cast Boolean columns to `np.uint8` before computing the range.

Both fixes align the pandas path with the policy already established in this codebase: `pandas.api.types.is_numeric_dtype(bool)` is `True`. After this change, `clinical_kernel(df)` and `clinical_kernel(df.astype({col: 'uint8'}))` produce identical kernel matrices, and `ClinicalKernelTransform().fit(df)` no longer crashes.

**Tests**

Added `TestClinicalKernel.test_bool_column_treated_as_numeric` asserting:
- `clinical_kernel` returns the same matrix for a Boolean column and its `uint8` equivalent.
- `ClinicalKernelTransform().fit` produces matching `_numeric_ranges` and `X_fit_` for both dtypes.
- The Boolean column is classified into `_numeric_columns`, not `_nominal_columns`.

Local verification:
- `pytest tests/` (including slow tests): 966 passed, 48 skipped, 0 failed.
- `pytest --doctest-modules --pyargs sksurv.kernels`: 1 passed.
- `ruff check` on the whole repo: clean.
- `pre-commit run` on changed files: all hooks pass.

**Behavior change note**

Users whose pipelines previously called `clinical_kernel` on a pandas frame containing Boolean columns will see different (corrected) kernel values after this change. The previous values were silently biased, so the change is a bug fix rather than an API change, but it may be worth flagging in the changelog.

### DIFF
--- a/sksurv/kernels/clinical.py
+++ b/sksurv/kernels/clinical.py
@@ -12,7 +12,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import numpy as np
 import pandas as pd
-from pandas.api.types import CategoricalDtype, is_numeric_dtype
+from pandas.api.types import CategoricalDtype, is_bool_dtype, is_numeric_dtype
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import _check_feature_names, _check_n_features, check_is_fitted
 
@@ -39,7 +39,7 @@ def _get_continuous_and_ordinal_array(x):
     """Convert array from continuous and ordered categorical columns"""
     nominal_columns = x.select_dtypes(include=["object", "category"]).columns
     ordinal_columns = pd.Index([v for v in nominal_columns if x[v].cat.ordered])
-    continuous_columns = x.select_dtypes(include=[np.number]).columns
+    continuous_columns = x.select_dtypes(include=[np.number, "bool"]).columns
 
     x_num = x.loc[:, continuous_columns].to_numpy(dtype=np.float64)
     if len(ordinal_columns) > 0:
@@ -205,6 +205,8 @@ class ClinicalKernelTransform(BaseEstimator, TransformerMixin):
 
                 col = col.cat.codes
             elif is_numeric_dtype(dt):
+                if is_bool_dtype(dt):
+                    col = col.astype(np.uint8)
                 numeric_ranges.append(col.max() - col.min())
                 numeric_columns.append(i)
             else:
--- a/tests/test_clinical_kernel.py
+++ b/tests/test_clinical_kernel.py
@@ -319,6 +319,25 @@ Feature names unseen at fit time:
             t.prepare(data)
 
     @staticmethod
+    def test_bool_column_treated_as_numeric():
+        df_bool = pd.DataFrame(
+            {
+                "age": [20.0, 23.0, 26.0, 54.0, 100.0],
+                "event": [True, False, True, True, False],
+            }
+        )
+        df_uint = df_bool.assign(event=df_bool["event"].astype(np.uint8))
+
+        assert_array_almost_equal(clinical_kernel(df_bool), clinical_kernel(df_uint))
+
+        t_bool = ClinicalKernelTransform().fit(df_bool)
+        t_uint = ClinicalKernelTransform().fit(df_uint)
+        assert_array_almost_equal(t_bool._numeric_ranges, t_uint._numeric_ranges)
+        assert_array_almost_equal(t_bool.X_fit_, t_uint.X_fit_)
+        assert list(t_bool._numeric_columns) == [0, 1]
+        assert list(t_bool._nominal_columns) == []
+
+    @staticmethod
     def test_feature_mismatch(make_data):
         data, _ = make_data()
         x = data.iloc[:, :2]


sebp (approved):
Thanks a lot for this!

PR #592, opened 2026-05-25
Add support for polars DataFrame

### LINKED ISSUE #591: ENH: support polars DataFrame as input and output
### Feature Type

- [x] Adding new functionality to scikit-survival

- [ ] Changing existing functionality in scikit-survival

- [ ] Removing existing functionality in scikit-survival

### Problem Description

scikit-survival currently accepts only pandas DataFrames as input. Users whose training pipelines are built on polars must convert to pandas at the boundary (`df.to_pandas()`) before calling scikit-survival estimators, transformers, kernels, or dataset loaders. The conversion costs memory and runtime, defeats polars' lazy execution, and forces users to maintain a dual representation in real-world ML pipelines.

While investigating how to add native polars support, I noticed that scikit-learn 1.4+ has been gradually adopting [narwhals](https://narwhals-dev.github.io/narwhals/) as its dataframe-agnostic compatibility layer (see [scikit-learn/scikit-learn#31127](https://github.com/scikit-learn/scikit-learn/issues/31127) and related PRs). narwhals is a thin, dependency-free shim that exposes a single API surface for pandas, polars, pyarrow, modin, cuDF, and others.

Adopting the same approach in scikit-survival has two benefits:

1. **Immediate**: native polars `DataFrame` / `LazyFrame` / `Series` support.
2. **Future**: once the boundary layer is in place, supporting another Narwhals-backed library (pyarrow Table, modin, cuDF, ...) is a matter of extending a single registry rather than touching every estimator.

This keeps scikit-survival aligned with where scikit-learn itself is heading, and avoids tying the codebase to a single non-pandas dataframe library.

### Feature Description

Accept `polars.DataFrame`, `polars.LazyFrame`, and `polars.Series` as inputs to estimators, transformers, kernels, IO, and dataset loaders, alongside the existing pandas inputs. Dataset loaders and `loadarff` gain an `output_type="polars"` option so an entire pipeline can stay in polars.

```python
import polars as pl
from sksurv.datasets import load_whas500
from sksurv.linear_model import CoxPHSurvivalAnalysis

X, y = load_whas500(output_type="polars")
CoxPHSurvivalAnalysis().fit(X, y).predict(X)
```

Internally, a new boundary module (`sksurv._dataframe`) routes non-pandas inputs through `narwhals.stable.v2`. The pandas code paths are unchanged, so existing users see no behavioral difference. Adding another Narwhals-supported backend later requires only extending the boundary module's library registry.

I have already prepared a working implementation following this design and can submit it as a pull request.

### Alternative Solutions

- **Call `df.to_pandas()` at the boundary**: forfeits polars' zero-copy and lazy semantics, doubles memory for large frames, and forces every downstream user to write conversion boilerplate.
- **Polars-only support via direct `import polars`**: would work for polars but lock the codebase into a second hard dataframe dependency, and offer no path for pyarrow / modin / cuDF later. This is the path scikit-learn explicitly chose to avoid by going through narwhals.

### References and existing implementations

- [narwhals](https://narwhals-dev.github.io/narwhals/) — dataframe-agnostic compatibility layer used by scikit-learn, altair, plotly, and others.
- [scikit-learn/scikit-learn#31127](https://github.com/scikit-learn/scikit-learn/issues/31127) — scikit-learn's ongoing narwhals adoption.
- scikit-learn 1.4+ already accepts polars output via `set_output(transform="polars")`. This issue extends that direction by also accepting polars *input* in scikit-survival.

55Kamiryo:
**Checklist**

* [x] Closes #591
* [x] `pytest` passes
* [x] Tests are included
* [x] Code is well formatted

**What does this implement or fix? Explain your changes.**

This PR adds support for `polars.DataFrame` inputs across the main dataframe entry points in estimators, kernels, preprocessing, I/O, and dataset loaders.

It also adds `polars.Series` support where the existing public API already accepts single-column inputs, such as column utilities.

`polars.LazyFrame` is intentionally not supported. It is rejected at the boundary with a clear error asking the caller to call `.collect()` first.

Existing pandas behavior is intended to remain unchanged.

Internally, this introduces a private `sksurv._dataframe` boundary built on top of [Narwhals](https://narwhals-dev.github.io/narwhals/). For the column, preprocessing, and clinical-kernel dataframe paths touched here, pandas and polars now share a single narwhals-backed implementation rather than mirrored, library-specific code paths.

`PandasDataFrameLibrary` and `PolarsDataFrameLibrary` carry only what narwhals cannot express directly:

* type predicates
* ordered-categorical detection
* backend-specific cast-error detection

The goal is to centralize dataframe-library-specific handling so that future support for additional narwhals-compatible dataframe libraries can be explored without adding per-estimator dispatch logic.

See #591 for the motivation.

### Highlights

* Main estimator `fit`, `predict`, `score`, and related prediction paths now accept numeric `polars.DataFrame` inputs.
* `polars.LazyFrame` is rejected at the boundary via `ensure_eager_dataframe` / `to_narwhals_dataframe`, with a clear error asking the caller to call `.collect()` first.
* Dataset loaders, including `load_whas500` and `load_gbsg2`, as well as `loadarff` and `writearff`, now support `output_type="polars"`.
* `OneHotEncoder` and `ClinicalKernelTransform` support polars inputs, while requiring `fit` and `transform` to use the same dataframe library. For example, fitting with pandas and transforming with polars, or vice versa, raises a `TypeError`, consistent with scikit-learn conventions.
* Existing pandas behavior is preserved and covered by parity tests that check pandas and polars against the same expectations.
* `sksurv._dataframe` centralizes dataframe-library-specific handling and is intended to be the extension point for future narwhals-backed dataframe libraries.

### Scope considerations

I'd appreciate your guidance on scope after you review the diff. I'm happy to expand this PR or split items into follow-ups, depending on what you prefer.

#### scikit-learn 1.9 migration

This branch currently also carries the scikit-learn 1.9 migration, but #593 already does that and the two PRs overlap heavily.

I'm leaning toward letting #593 own the 1.9 migration, then rebasing this PR on top once #593 lands. See my comment below for details.

#### Items I'm currently leaning toward leaving out of this PR

* **Native polars implementation of `writearff`.** The current implementation converts polars input to pandas column by column before serializing. This is correct, but not zero-copy. A native implementation seems better suited to a separate change with its own focused tests.
* **Consolidating `test_polars_*.py` with the corresponding pandas tests via `@pytest.mark.parametrize("backend", ...)`.** This is mostly a mechanical refactor. Now that the touched paths share a single implementation, it could be handled as a focused follow-up.

#### Possible follow-ups

* Validate and enable additional narwhals-supported backends, such as PyArrow `Table`, Modin, and cuDF, by extending `DATAFRAME_LIBRARIES` in `sksurv/_dataframe/_input.py` and adding parametrized parity tests. The boundary layer is designed to minimize per-estimator changes, but I'd like to confirm that against at least one additional backend before claiming broader support.
* Revisit whether `output_type` should accept an `nw.Implementation`, or a string alias resolved through it, rather than a fixed literal. That would avoid growing the public API each time a backend is enabled. I'm not yet sure that is the right interface.

I'm happy to bring any of these into this PR if you'd like the scope to be wider, or to file separate issues or follow-up PRs.


sebp:
Thanks for getting this started.

I haven't looked at what scikit-learn 1.9 provides now, but we should re-use as much as possible. I don't mind accessing sklearn private API. For instance, sklearn's `validate_data` is being replaced by the new `validate_data_with_eager_dataframe`. What is the reason for it?

In the API doc examples, comments about polars usage have been added. I think this is not necessary.

For the tests, rather than duplicating the data, can we use a single source for the data use the data frame interchange protocol to convert to the desired data frame library?

The `safe_concat` function could actually be obsolete now. Originally, it was a work-around for pandas, but I think it has been fixed for a while. 

I believe scikit-learn supports pyarrow tables too. Might be something to consider for later.

55Kamiryo:
@sebp 
Thanks for the review. A short rundown of how each point was addressed:

- **`validate_data_with_eager_dataframe`** — wrapper removed. Estimators now call scikit-learn's `validate_data(self, collect_lazy_dataframe(X), ...)` directly. I kept `collect_lazy_dataframe` because I verified that neither sklearn 1.8 nor 1.9 accepts a polars `LazyFrame` through `validate_data` (`narwhals.dependencies.is_into_dataframe` excludes lazy frames by design).
- **Polars comments in API doc examples** — removed.
- **Test data duplication** — added `tests/dataframe_test_utils.py` with shared pandas factories and a `to_polars_via_interchange()` helper. Parity tests now pull from the same source. A few direct `pl.DataFrame(...)` constructions remain on purpose, for polars-specific semantics (Enum / Categorical / LazyFrame) that don't round-trip cleanly through the interchange protocol.
- **`safe_concat`** — removed entirely. As you noted, pandas has long since fixed the categorical-preservation issue it was originally working around, so the function is no longer needed. The only internal caller (`datasets/base.py`) now uses a small local helper that dispatches to `pd.concat` / `nw.concat`.
- **PyArrow** — the boundary layer keeps a single `EXTERNAL_DATAFRAME_LIBRARIES` tuple, so adding PyArrow later is a `_pyarrow.py` module plus one tuple entry. I'd prefer to validate it in a follow-up PR with parity tests rather than claim support here.

sksurv/_dataframe/_categorical_encoding.py
```diff
@@ -0,0 +1,91 @@
...
+            continue
+
+        encoded = column_to_one_hot_matrix(col, semantics, drop_first=True)
+        new_names = get_one_hot_column_names(semantics, drop_first=True)
+        # Build the one-hot block in a single backend call so polars stays on
+        # its zero-copy numpy path instead of round-tripping through Python lists.
+        encoded_frame = nw.from_numpy(encoded, schema=list(new_names), backend=implementation)
+        output_frames.append(encoded_frame)
+
+    if not output_frames:
```
sebp: be explicit here and use `len(output_frames) == 0`

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
...
+            output_frames.append(new_col.to_frame())
+        else:
+            output_frames.append(nw_table.select(col_name))
+
+    return nw.concat(output_frames, how="horizontal").to_native()
+
+
+def encode_categorical_narwhals(table, columns=None, allow_drop=True):
+    library = get_dataframe_library(table, allow_series=True)
+    if library is not None and get_dataframe_library(table) is None:
```
sebp: What does this condition check? I suppose whether table is a data frame or series. It's not obvious though

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+
+def _prepare_column_for_one_hot(col):
+    dtype = col.dtype
+    if isinstance(dtype, nw.Boolean):
+        # Match pandas get_dummies(drop_first=True): False is the dropped baseline.
+        col = col.replace_strict({True: "True", False: "False"}, default=None, return_dtype=nw.String)
+        observed = sorted({"True", "False"} & set(col.drop_nulls().unique().to_list()))
+        observed = sorted(observed, key=lambda v: v != "False")
```
sebp: Why is `sorted` being called twice? Is it a complicated way to reverse the sort order?

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
...
+
+    encoded = column_to_one_hot_matrix(nw_series, semantics, drop_first=True)
+    new_names = get_one_hot_column_names(semantics, drop_first=True)
+    # Single backend construction keeps polars on its zero-copy numpy path.
+    return nw.from_numpy(encoded, schema=list(new_names), backend=implementation).to_native()
+
+
+def categorical_to_numeric_narwhals(table):
+    library = get_dataframe_library(table, allow_series=True)
+    if library is not None and get_dataframe_library(table) is None:
```
sebp: Same as above, it's not intuitive what this is doing.

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
...
+def _encode_series_as_numeric_codes(nw_series):
+    dt = nw_series.dtype
+    if isinstance(dt, nw.Boolean):
+        return nw_series.cast(nw.Int64).to_numpy()
+    if isinstance(dt, nw.String):
+        try:
+            return nw_series.cast(nw.Int64).to_numpy()
+        except Exception as exc:
+            if not is_non_numeric_cast_error(exc):
+                raise
```
sebp, line 204: This is a strange statement. What is it trying to achieve?

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
...
+    semantics = infer_column_semantics(nw_series)
+    if semantics.kind == "numeric":
+        return nw_series.to_numpy()
+    if semantics.kind == "nominal" and isinstance(dt, (nw.Categorical, nw.String)):
+        categories = tuple(sorted(semantics.categories or ()))
+    else:
+        categories = semantics.categories or ()
```
sebp: When would this condition be triggered? I suppose numeric values have no categories

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,186 @@
+"""Dataframe implementations for :mod:`sksurv.column` functions."""
```
sebp, line 13: Please add license header.

sksurv/_dataframe/_categorical_semantics.py
```diff
@@ -0,0 +1,63 @@
+"""Column semantics for Narwhals-backed dataframe inputs.
```
sebp, line 13: Please add license header.

sksurv/_dataframe/_categorical_encoding.py
```diff
@@ -0,0 +1,91 @@
+"""Semantics-based dataframe encoding helpers."""
```
sebp, line 13: Please add license header.

sksurv/_dataframe/__init__.py
```diff
@@ -0,0 +1,41 @@
+"""Internal dataframe helpers for Narwhals-backed input handling."""
```
sebp, line 13: Please add license header.

sksurv/_dataframe/_input.py
```diff
@@ -0,0 +1,102 @@
...
+SUPPORTED_DATAFRAME_INPUT_TYPES = _oxford_join(
+    ["pandas.DataFrame"] + [name for lib in EXTERNAL_DATAFRAME_LIBRARIES for name in lib.dataframe_display_names]
+)
+
+
+def get_dataframe_library(obj, *, allow_series=False):
+    for library in EXTERNAL_DATAFRAME_LIBRARIES:
+        if library.is_dataframe(obj) or (allow_series and library.is_series(obj)):
+            return library
+    return None
```
sebp: Line can be removed

sksurv/_dataframe/_input.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def is_supported_dataframe(obj):
+    """Return whether ``obj`` is a dataframe type explicitly supported by sksurv."""
+    return nw.dependencies.is_pandas_dataframe(obj) or get_dataframe_library(obj) is not None
```
sebp: Instead of checking pandas explicitly, can we add it to `EXTERNAL_DATAFRAME_LIBRARIES`?

sksurv/_dataframe/_polars.py
```diff
@@ -0,0 +1,59 @@
...
+
+__all__ = ["LIBRARY", "PolarsDataFrameLibrary", "is_dataframe", "is_series"]
+
+
+def is_dataframe(obj):
+    return nw.dependencies.is_polars_dataframe(obj) or nw.dependencies.is_polars_lazyframe(obj)
+
+
+def is_series(obj):
+    return nw.dependencies.is_polars_series(obj)
```
sebp: Can these be moved directly to the class below?

sksurv/io/arffread.py
```diff
@@ -43,18 +43,60 @@ def _to_pandas(data, meta):
     return pd.DataFrame.from_dict(data_dict)
 
 
-def loadarff(filename):
+def _decode_nominal_value(b):
+    if b == b"?":
+        return None
+    return b.decode()
```
sebp: Can be simplified to

```python
if b != b"?":
    return b.decode()
```

sksurv/io/arffread.py
```diff
@@ -43,18 +43,60 @@ def _to_pandas(data, meta):
...
+        Dataframe library for the returned frame. Nominal columns become
+        ``pd.Categorical(ordered=False)`` (pandas) or
+        ``pl.Enum(declared_categories)`` (polars), preserving the full
+        declared category list including labels absent from the data.
 
     Returns
     -------
-    data_frame : :class:`pandas.DataFrame`
-        DataFrame containing data of ARFF file
+    data_frame : :class:`pandas.DataFrame` or :class:`polars.DataFrame`
```
sebp, line 98: Add polars to `intersphinx_mapping` in `doc/conf.py`

sksurv/io/arffwrite.py
```diff
@@ -55,13 +83,12 @@ def writearff(data, filename, relation_name=None, index=True):
     ...     'class': ['A', 'B', 'C']
     ... }, index=['One', 'Two', 'Three'])
     >>>
-    >>> # Write to ARFF file
-    >>> writearff(data, 'test_output.arff', relation_name='test_data')
+    >>> # Write to ARFF file (use a temporary path so the CWD stays clean)
+    >>> out_path = Path(tempfile.mkdtemp()) / 'out.arff'
+    >>> writearff(data, str(out_path), relation_name='test_data')
```
sebp: Use

```python
with tempfile.NamedTemporaryFile(suffix=".arff", delete_on_close=False) as temp:
    writearff(data, temp.name, relation_name='test_data')
```

sksurv/io/arffwrite.py
```diff
@@ -36,14 +61,17 @@ def writearff(data, filename, relation_name=None, index=True):
         Name of relation in ARFF file.
 
     index : boolean, optional, default: True
-        Write row names (index)
+        Write row names (index). Ignored for polars input (polars has no
```
sebp: I guess it is only relevant for pandas, no other data frame library I'm aware of has an index.

sksurv/kernels/_clinical_dataframe.py
```diff
@@ -0,0 +1,277 @@
...
+    return x_numeric
+
+
+def _continuous_range(values):
+    return (values.max() - values.min()) if values.size > 0 else np.nan
+
+
+def _ordinal_range(codes):
+    valid = ~np.isnan(codes)
+    return (codes[valid].max() - codes[valid].min()) if np.any(valid) else 0.0
```
sebp: `np.nanmax` and `np.nanmin` can be used here

sksurv/kernels/_clinical_dataframe.py
```diff
@@ -0,0 +1,277 @@
...
+        if is_narwhals_dataframe(y):
+            raise TypeError("x and y must use the same dataframe library")
+        if not x.columns.equals(y.columns):
+            raise ValueError("columns do not match")
+
+
+def _normalize_nominal_nulls_for_kernel(arr):
+    if arr.dtype != object:
+        return arr
+    mask = np.frompyfunc(lambda v: v is None, 1, 1)(arr).astype(bool)
```
sebp: wouldn't `mask = arr == None` achieving the same?

sksurv/kernels/clinical.py
```diff
@@ -160,8 +170,17 @@ class ClinicalKernelTransform(BaseEstimator, TransformerMixin):
            Annual International Conference of the IEEE Engineering in Medicine and Biology Society, 5913-7, 2009
     """
 
-    def __init__(self, *, fit_once=False, _numeric_ranges=None, _numeric_columns=None, _nominal_columns=None):
+    def __init__(
+        self,
+        *,
+        fit_once=False,
+        ordinal_columns=None,
```
sebp: Add a `_`prefix

sksurv/kernels/clinical.py
```diff
@@ -160,8 +170,17 @@ class ClinicalKernelTransform(BaseEstimator, TransformerMixin):
...
+        _numeric_ranges=None,
+        _numeric_columns=None,
+        _nominal_columns=None,
+    ):
         self.fit_once = fit_once
+        self.ordinal_columns = ordinal_columns
```
sebp: Add a `_` prefix

sksurv/metrics.py
```diff
@@ -737,6 +740,9 @@ def integrated_brier_score(survival_train, survival_test, estimate, times):
     Load and prepare data.
 
     >>> X, y = load_gbsg2()
+    >>> # ``.map(len)`` converts the roman-numeral tumor grade to an integer
+    >>> # ordinal. Polars equivalent::
+    >>> #     X = X.with_columns(pl.col('tgrade').cast(pl.String).str.len_chars().cast(pl.Int64))
```
sebp: Comment seems to be outdated.

sksurv/metrics.py
```diff
@@ -612,6 +612,9 @@ def brier_score(survival_train, survival_test, estimate, times):
...
```
sebp: Comment seems to be outdated.

sebp (changes requested):
Thanks for the quick turnaround.

I'm half way through reviewing the code, but still want to share my comments already.

## pandas vs polars code paths

One general concern is that now there are several instances where there is a regular pandas path and a narwhals path, e.g.:
```python
def standardize(table, with_std=True):
    if is_narwhals_dataframe(table):
        return standardize_narwhals_dataframe(table, with_std=with_std)
    return _apply_along_column(table, standardize_column, with_std=with_std)
```

Ideally, everything would be handled by narwhals, but I guess narwhals can't account for everything, like ordinal dtypes being absent in polars. An alternative could be to extend the `_dataframe` module with a `PandasDataFrameLibrary`and add the respective implementations there. Then, the call would simply be `library_module.standardize` instead of an if-else statement. Unfortunately, things get even more complicated when inputs can also be numpy arrays.

## LazyFrame

If sklearn does not support LazyFrame, I think there is little benefit supporting it in sksurv, because one still could not use pipelines with preprocessing from sklearn and an estimator from sksurv. I think it is reasonable to ask users to call `.collect()` themselves. Then the question is how difficult it is with narwhals to distinguish between regular and lazy frames.

55Kamiryo:
Thanks for the detailed review. I've pushed changes addressing the inline comments, with notes in the two threads where I kept the parameter public.

A few notes on the higher-level points:

## pandas vs. polars code paths

I'd initially planned to keep the existing pandas paths separate and converge them in a follow-up, to keep this PR focused on adding polars support.

Since you pointed at `PandasDataFrameLibrary`, I took that direction and pushed the unification further here. I'm happy to split the convergence back out into a follow-up if you'd rather keep this PR narrower.

For the column, preprocessing, and clinical dataframe paths touched by this review, pandas and polars now share narwhals-backed implementations instead of mirrored pandas-specific dataframe paths.

`PandasDataFrameLibrary` and `PolarsDataFrameLibrary` now only carry what narwhals cannot express directly:

* type predicates
* ordered-categorical detection
* backend-specific cast-error detection

The dataframe-vs-**numpy** branches that remain are the numpy case you flagged as the hard one. I'm happy to fold those into the abstraction too if you'd prefer, though numpy is not a dataframe library, so it does not fit the interface as cleanly.

## LazyFrame

Agreed — I dropped `LazyFrame` support.

`polars.LazyFrame` is now rejected at the boundary via `ensure_eager_dataframe` / `to_narwhals_dataframe`, with:

```text
TypeError: polars.LazyFrame is not supported; call .collect() before passing to scikit-survival.
```

This is covered at every entry point. Distinguishing lazy from eager is a one-liner with `nw.dependencies.is_polars_lazyframe`.

I'll also drop `LazyFrame` from the PR title.

## Test data conversion

I moved away from the dataframe interchange protocol after checking the current behavior.

Polars has deprecated `DataFrame.__dataframe__()`, and that path drops the declared `Enum` categories that the parity tests rely on. As a result, the shared helper, `tests/dataframe_test_utils.py::to_polars_dataframe`, now builds the polars frame directly.

I'm happy to switch back to interchange if you'd prefer, despite the `Enum` category loss.

## Scope: scikit-learn 1.9 / overlap with #593

This PR currently also carries the scikit-learn 1.9 migration in its own commit, but I see that #593 already does that, and the two overlap almost entirely:

* `ensemble/boosting.py`
* `ensemble/forest.py`
* `tree/_criterion.pyx`
* `tree/tree.py`
* `test_forest`
* `test_stacking`
* CI dependency pins
* boosting user-guide outputs

#593 is also cleaner in places — for example, it drops the `criterion` parameter outright per scikit-learn#32708, whereas I only changed its default.

Rather than duplicate that work, I'd suggest letting #593 own the 1.9 migration and keeping this PR focused on polars/narwhals support.

The polars handling here is written against 1.9, for example with `validate_data(..., skip_check_array=True)`, so it naturally sits on top. Once #593 lands, I'll rebase #592 onto `main` and drop my 1.9 commits.

I'll keep carrying the migration here until then, so this PR is not blocked if #593 takes a while.

Does that sequencing work — merge #593 first, then I rebase #592 on top?

If you'd rather not couple the two, I'm happy to keep the 1.9 migration here instead and align it with #593's approach, for example by dropping `criterion` outright rather than only changing its default.

Likewise for the pandas-to-narwhals convergence, I'm happy to split or realign the scope however you prefer.

sksurv/kernels/clinical.py
55Kamiryo: I kept this public, renaming `ordinal_columns` to `ordinal_categories`, rather than adding the `_` prefix, because this is a user-facing parameter rather than internal state.

It is the only way to declare which columns are ordinal and what their category order is. This is needed for polars, whose `Enum` / `Categorical` types do not carry an `ordered` flag. Pandas `Categorical(ordered=True)` is still auto-detected.

This also mirrors the functional API, `clinical_kernel(..., ordinal_categories=...)`, and is covered by the parity tests in `test_polars_clinical_kernel.py`.

The underscore-prefixed parameters next to it — `_numeric_ranges`, `_numeric_columns`, and `_nominal_columns` — are a different kind of parameter: internal fitted state that is round-tripped through `__init__` for the `fit_once` / `prepare()` path. That is probably what made `ordinal_columns` look out of place next to them.

I renamed it to `ordinal_categories` and changed it to a `{name: [ordered labels]}` mapping to make the user-facing intent clearer.

I'm happy to move ordinal declaration elsewhere, for example to a dedicated method, if you'd prefer not to expose it through `__init__`.

sksurv/kernels/clinical.py
55Kamiryo: Same reasoning as on line 177 — I kept this public as `ordinal_categories` rather than prefixing it with _. See my reply there.

sksurv/io/arffwrite.py
55Kamiryo: I used `tempfile.TemporaryDirectory()` instead.

`NamedTemporaryFile`'s `delete_on_close` argument is only available on Python 3.12+, while this project still supports Python 3.11 via `requires-python >=3.11`. As a result, the suggested snippet would fail the doctest on the 3.11 CI job.

`TemporaryDirectory` keeps the example self-cleaning while working on all supported Python versions. It also fixes the original `mkdtemp()` usage, which left a directory behind.

sksurv/_dataframe/_input.py
55Kamiryo: Done — pandas is now a first-class entry in the dataframe libraries tuple, so the predicate dispatches through the same library abstraction instead of checking pandas explicitly.

I also renamed `EXTERNAL_DATAFRAME_LIBRARIES` to `DATAFRAME_LIBRARIES`, since pandas is no longer special-cased.

sksurv/_dataframe/_categorical_encoding.py
55Kamiryo: Done — now if len(output_frames) == 0:.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done — removed the double sort; now an explicit ("False", "True") ordering with a comment on why the order matters for drop_first.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done — replaced the opaque two-call condition with a named helper, is_supported_series().

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done — replaced the opaque two-call condition with a named helper, is_supported_series().

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done — added a comment: re-raise anything other than the swallowed non-numeric-cast error.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Right — numeric is handled by the callers, so this is reached only for nominal columns. Removed the dead numeric branch and added a comment.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done — added the GPLv3 header.

sksurv/_dataframe/_categorical_semantics.py
55Kamiryo: Done — added the GPLv3 header.

sksurv/_dataframe/_categorical_encoding.py
55Kamiryo: Done — added the GPLv3 header.

sksurv/_dataframe/__init__.py
55Kamiryo: Done — added the GPLv3 header.

sksurv/_dataframe/_input.py
55Kamiryo: Done — removed.

sksurv/_dataframe/_polars.py
55Kamiryo: Done — moved into PolarsDataFrameLibrary as staticmethods.

sksurv/io/arffread.py
55Kamiryo: Done

sksurv/io/arffread.py
55Kamiryo: Done

sksurv/io/arffwrite.py
55Kamiryo: Done — scoped the index note to pandas.

sksurv/kernels/_clinical_dataframe.py
55Kamiryo: Done — now np.nanmax / np.nanmin in _ordinal_range.

sksurv/kernels/_clinical_dataframe.py
55Kamiryo: Done — now mask = arr == None (with a noqa: E711 and a comment, since it's an element-wise object-array check).

sksurv/metrics.py
55Kamiryo: Done — updated the stale comment.

sksurv/metrics.py
55Kamiryo: Done — updated the stale comment.

sebp:
@55Kamiryo Thanks for the updates, this is great work.

I merged https://github.com/sebp/scikit-survival/pull/593 now, which adds support for scikit-learn 1.9. Please me let me know if you can rebase your change on top of it.

55Kamiryo:
@sebp 
Thank you for the review.

I've rebased this PR onto `main`, so it now builds on your scikit-learn 1.9 work. I dropped my own 1.9 commits since #593 already covers that. CI is green.

Since your last review, I also added a few focused tests and made small fixes around docstrings and edge cases. The substantive behavior is unchanged from the version you reviewed.

This is a fairly large change, so I'll keep an eye on it for any issues that come up.

I'll keep the remaining follow-ups separate, including additional backends such as PyArrow, error handling, and the `LazyFrame` boundary behavior. I also opened scikit-learn/scikit-learn#34258 to make `check_array` report a clearer error for polars `LazyFrame` inputs.

Happy to adjust anything you'd like.

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+    codes = int_series.fill_null(-1).to_numpy()
+    if np.any(null_mask):
+        codes = codes.astype(float)
+        codes[null_mask] = np.nan
+    return codes
+
+
+def _encode_series_as_numeric_codes(nw_series):
+    dt = nw_series.dtype
+    if isinstance(dt, nw.Boolean):
```
sebp: Can be `dt.is_boolean()`

sksurv/_dataframe/_categorical_encoding.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Narwhals rebuilds columns positionally, so values reconstructed from numpy
+    lose the caller's pandas index. Capture it here and restore it afterwards
+    with :func:`reattach_pandas_index` so the returned pandas frame keeps the
+    caller's original row labels. No-op for index-less backends (e.g. polars),
+    which yield ``None``.
+    """
+    if nw_obj.implementation.is_pandas_like():
+        return nw_obj.to_native().index
+    return None
```
sebp: Line can be removed, default is to return `None`.

sksurv/_dataframe/_categorical_encoding.py
```diff
@@ -0,0 +1,147 @@
...
+    source index is non-default or non-unique. Resetting the source index up
+    front forces positional alignment; restore it afterwards with
+    :func:`reattach_pandas_index`. No-op for index-less backends (e.g. polars).
+
+    Returns a ``(frame, original_index)`` pair; ``original_index`` is ``None``
+    for non-pandas backends.
+    """
+    original_index = capture_pandas_index(nw_frame)
+    if original_index is not None:
+        return nw.from_native(nw_frame.to_native().reset_index(drop=True)), original_index
```
sebp, line 58: Is it really necessary to create a new Narwhals data frame without the index? I thought Narwhals operations always work positionally (like polars), but keep track of the pandas index too.

sksurv/_dataframe/_categorical_semantics.py
```diff
@@ -0,0 +1,81 @@
...
+    categories: tuple | None
+    ordered: bool
+
+
+def infer_column_semantics(column):
+    """Infer :class:`ColumnSemantics` from a single column."""
+    s = nw.from_native(column, series_only=True)
+    dt = s.dtype
+
+    if dt.is_numeric() or isinstance(dt, nw.Boolean):
```
sebp: `dt.is_boolean()` can be used.

sksurv/_dataframe/_categorical_semantics.py
```diff
@@ -0,0 +1,81 @@
...
+        return ColumnSemantics(name=s.name, kind="numeric", categories=None, ordered=False)
+
+    if isinstance(dt, (nw.Enum, nw.Categorical, nw.String, nw.Object)):
```
sebp: This type of statement appears several times. It might be useful to extract a function for.

```
sksurv/_dataframe/_column_impl.py
68:        if not isinstance(dt, (nw.Categorical, nw.Enum, nw.String, nw.Object)):
83:            if isinstance(dtype, (nw.Categorical, nw.Enum, nw.String, nw.Object))

sksurv/_dataframe/_categorical_semantics.py
73:    if isinstance(dt, (nw.Enum, nw.Categorical, nw.String, nw.Object)):

sksurv/preprocessing.py
180:            if isinstance(dtype, (nw.Categorical, nw.Enum, nw.String, nw.Object))

sksurv/kernels/_clinical_dataframe.py
89:        if not isinstance(dtype, (nw.Categorical, nw.Enum, nw.String, nw.Object)):
102:    if not isinstance(dtype, (nw.Categorical, nw.Enum, nw.String, nw.Object)):
202:    if not isinstance(dtype, (nw.Categorical, nw.Enum, nw.String, nw.Object)):
```

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+
+def _prepare_column_for_one_hot(col):
+    dtype = col.dtype
+    if isinstance(dtype, nw.Boolean):
+        # Match pandas get_dummies(drop_first=True): False is the dropped baseline.
+        col = col.replace_strict({True: "True", False: "False"}, default=None, return_dtype=nw.String)
+        present = set(col.drop_nulls().unique().to_list())
+        # Keep "False" before "True" so drop_first drops "False" as the baseline.
+        observed = [value for value in ("False", "True") if value in present]
+        return col, ColumnSemantics(name=col.name, kind="nominal", categories=tuple(observed), ordered=False)
```
sebp, line 108: Is it really necessary to convert everything to string?

Looking at https://github.com/55Kamiryo/scikit-survival/blob/3e073ad40935fc96a814e348faf7cdbc03e7cf7b/sksurv/_dataframe/_categorical_encoding.py#L76

I think setting categories to `(False, True)`, or the value actual present, achieve the same

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+        allow_drop=allow_drop,
+        implementation=implementation,
+        on_empty="empty_frame",
+        logger=logging.getLogger("sksurv"),
+    )
+
+
+def _prepare_column_for_one_hot(col):
+    dtype = col.dtype
+    if isinstance(dtype, nw.Boolean):
```
sebp: Can use `dtype.is_boolean()`

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+        result = nw.from_dict({}, backend=implementation).to_native()
+        if original_index is not None:
+            result = result.reindex(original_index)
+        return result
+    if not allow_drop and n_cat <= 1:
+        return nw_series.to_native()
+
+    encoded = column_to_one_hot_matrix(nw_series, semantics, drop_first=True)
+    new_names = get_one_hot_column_names(semantics, drop_first=True)
+    result = nw.from_numpy(encoded, schema=list(new_names), backend=implementation).to_native()
```
sebp: Change `get_one_hot_column_names()` to return a list, then skip the list conversion

sksurv/_dataframe/_categorical_encoding.py
```diff
@@ -0,0 +1,147 @@
...
+            if logger is not None:
+                logger.warning(f"dropped categorical variable {col_name!r}, because it has only {n_cat} values")
+            continue
+        if not allow_drop and n_cat <= 1:
+            output_frames.append(nw_X.select(col_name))
+            continue
+
+        encoded = column_to_one_hot_matrix(col, semantics, drop_first=True)
+        new_names = get_one_hot_column_names(semantics, drop_first=True)
+        encoded_frame = nw.from_numpy(encoded, schema=list(new_names), backend=implementation)
```
sebp: Change `get_one_hot_column_names()` to return a list, then skip the list conversion

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+            continue
+        col = nw.col(col_name)
+        # Mean/std must skip missing values, but polars only skips null in
+        # aggregations and propagates float NaN, so compute the statistics on a
+        # NaN-to-null normalized expression. The subtraction uses the original
+        # column, so a present row keeps its standardized value and a row that is
+        # missing stays missing. A fully-missing column standardizes to null on
+        # polars and NaN on pandas; both become NaN via to_numpy(), which is how
+        # the data reaches the estimators.
+        stats_col = col.fill_nan(None) if isinstance(dtype, (nw.Float32, nw.Float64)) else col
```
sebp: `dtype.is_float()` can be used.

sksurv/_dataframe/_column_impl.py
```diff
@@ -0,0 +1,224 @@
...
+    new_names = get_one_hot_column_names(semantics, drop_first=True)
+    result = nw.from_numpy(encoded, schema=list(new_names), backend=implementation).to_native()
+    return reattach_pandas_index(result, original_index)
+
+
+def categorical_to_numeric_narwhals(table):
+    if is_supported_series(table):
+        nw_series = nw.from_native(table, series_only=True)
+        dt = nw_series.dtype
+        if dt.is_numeric() and not isinstance(dt, nw.Boolean):
```
sebp: `dt.is_boolean()` can be used.

sksurv/_dataframe/_input.py
```diff
@@ -0,0 +1,103 @@
...
+]
+
+DATAFRAME_LIBRARIES = (_pandas.LIBRARY, _polars.LIBRARY)
+
+
+def _oxford_join(items):
+    items = list(items)
+    if len(items) <= 2:
+        return " or ".join(items)
+    return ", ".join(items[:-1]) + ", or " + items[-1]
```
sebp: Line is not covered by tests, which is okay. Just add a `  # pragma: no cover` at the end of the line

sksurv/_dataframe/_input.py
```diff
@@ -0,0 +1,103 @@
...
+    collected before anything useful could happen. Rather than collecting
+    implicitly, callers are asked to materialize the frame explicitly via
+    ``.collect()``.
+    """
+    _reject_polars_lazyframe(obj)
+    return obj
+
+
+def to_narwhals_dataframe(obj):
+    _reject_polars_lazyframe(obj)
```
sebp: I believe this can be removed, because `ensure_eager_dataframe()`will be called before, which makes the entire function obsolete. Call `nw.from_native()` directly

sksurv/datasets/base.py
```diff
@@ -163,66 +209,78 @@ def load_arff_files_standardized(
...
 
     x_train, y_train = get_x_y(dataset, attr_labels, pos_label, survival)
 
     if path_testing is not None:
-        x_test, y_test = _load_arff_testing(path_testing, attr_labels, pos_label, survival)
+        x_test, y_test = _load_arff_testing(path_testing, attr_labels, pos_label, survival, output_type=output_type)
 
-        if len(x_train.columns.symmetric_difference(x_test.columns)) > 0:
+        train_cols = list(nw.from_native(x_train).columns)
+        test_cols = list(nw.from_native(x_test).columns)
```
sebp: Directly create two set object to avoid creating a list and later a set

sksurv/datasets/base.py
```diff
@@ -422,11 +511,16 @@ def load_breast_cancer():
         Patients in the TRANSBIG Multicenter Independent Validation Series."
         Clin. Cancer Res. 13(11), 3207–14 (2007)
     """
+    _validate_output_type(output_type)
```
sebp: Validation is also done in `loadarff()`. I would suggest to remove the check here, and in the other load_* functions

sksurv/io/arffread.py
```diff
@@ -39,22 +39,63 @@ def _to_pandas(data, meta):
...
+            values = [_decode_nominal_value(b) for b in data[name]]
+            columns.append(pl.Series(name, values, dtype=pl.Enum(list(attr_format))))
+        else:
+            arr = data[name]
+            if is_string_dtype(arr.dtype):
+                # scipy returns bytes for string-typed attributes; decode + null-map.
+                values = [_decode_nominal_value(b) for b in arr]
+                columns.append(pl.Series(name, values, dtype=pl.String))
+            else:
+                columns.append(pl.Series(name, arr))
```
sebp: Instead of having 3 append calls, I'd suggest to set `name`, `values`, `dtype` in each branch, and then do `columns.append(pl.Series(name, values, dtype=dtype))` once

sksurv/io/arffwrite.py
```diff
@@ -173,6 +219,9 @@ def to_str(x):
         return str(x)
 
     data = data.map(to_str)
+    # The header has already been written; coerce rows to object values so
+    # categorical-only frames can pass through ``_check_str_array``.
+    data = data.astype(object)
```
sebp, line 224: Wouldn't this make the first condition in `_write_attribute_categorical()` obsolete, because the dtype would never be categorical after this cast?

sksurv/kernels/_clinical_dataframe.py
```diff
@@ -0,0 +1,275 @@
...
+
+def _resolve_ordinal_categories(table, ordinal_categories):
+    """Merge user-declared ordinal categories with backend auto-detected ones.
+
+    Explicit ``ordinal_categories`` win. In addition, pandas ``Categorical``
+    columns declared with ``ordered=True`` are auto-detected as ordinal using
+    their declared order. Other backends (e.g. polars) contribute nothing
+    automatically.
+    """
+    resolved = dict(_normalize_ordinal_categories(ordinal_categories))
```
sebp: `_normalize_ordinal_categories()`already returns a dict

sksurv/kernels/clinical.py
```diff
@@ -27,41 +42,14 @@
 
 
 def _nominal_kernel(x, y, out):
-    """Number of features that match exactly"""
```
sebp, line 30: Please keep the comment

sksurv/kernels/clinical.py
```diff
@@ -246,19 +273,46 @@ def fit(self, X, y=None, **kwargs):  # pylint: disable=unused-argument
...
+            raise ValueError(f"expected 2d array, but got {ndim}")
 
-        _check_feature_names(self, X, reset=True)
-        _check_n_features(self, X, reset=True)
+        X = ensure_eager_dataframe(X)
+        validate_data(self, X, skip_check_array=True)
 
         if self.fit_once:
+            if is_supported_dataframe(X):
+                raise TypeError(
```
sebp, line 289: The message is slightly misleading, because `prepare()` returns nothing

sksurv/kernels/clinical.py
```diff
@@ -246,19 +273,46 @@ def fit(self, X, y=None, **kwargs):  # pylint: disable=unused-argument
...
+        out = np.empty((n_samples, n_features), dtype=np.float64)
+        for i, col_name in enumerate(nw_Y.columns):
+            col = nw_Y.get_column(col_name)
+            kind, semantics = self._fitted_categorical_semantics[i]
+            col_dtype = col.dtype
+            if col_dtype.is_numeric() or isinstance(col_dtype, nw.Boolean):
+                out[:, i] = col.to_numpy().astype(np.float64)
+                continue
+            if kind == "numeric":
+                out[:, i] = col.to_numpy().astype(np.float64)
```
sebp: can be combined into a single if block

sksurv/kernels/clinical.py
```diff
@@ -273,26 +327,37 @@ def transform(self, Y):
...
+        fit_impl = getattr(self, "_fit_implementation_", None)
+        if y_is_dataframe and fit_impl is not None and to_narwhals_dataframe(Y).implementation != fit_impl:
+            raise TypeError("fit and transform must use the same dataframe library")
 
         n_samples_x = self.X_fit_.shape[0]
 
-        Y = np.asarray(Y)
+        # Replay fit-time semantics if recorded; otherwise pass through
+        # (fit_once / prepare paths, or pre-encoded numeric input).
+        if getattr(self, "_fitted_categorical_semantics", None) is not None and y_is_dataframe:
```
sebp: use `hasattr` instead of `getattr`

tests/test_clinical_kernel.py
```diff
@@ -8,92 +9,13 @@
...
-        if with_nominal:
-            data_s["metastasis"] = pd.Categorical(data["metastasis"], categories=["no", "yes"], ordered=False)
-        expected = _get_expected_matrix(
+        return make_clinical_kernel_pandas_data(
             with_ordinal=with_ordinal, with_nominal=with_nominal, with_continuous=with_continuous
         )
 
-        return pd.DataFrame(data_s), expected
-
     return _make_data
```
sebp: directly return `make_clinical_kernel_pandas_data` without the wrapper `_make_data`

tests/test_io.py
```diff
@@ -155,6 +157,218 @@ def test_writearff(make_data_fn, temp_file, infer_string_context):
...
+    # SciPy's public loadarff rejects string attributes, but the converter still
+    # needs to handle the record-array shape SciPy uses for non-nominal fields.
+    class Meta:
+        @staticmethod
+        def names():
+            return ["note"]
+
+        def __getitem__(self, key):
+            assert key == "note"
+            return "string", None
```
sebp, line 266: Can you please explain a little bit more about this?

tests/test_polars_estimators.py
```diff
@@ -0,0 +1,514 @@
...
+            "ComponentwiseGradientBoostingSurvivalAnalysis",
+            lambda: ComponentwiseGradientBoostingSurvivalAnalysis(n_estimators=5, random_state=0),
+        ),
+        ("SurvivalTree", lambda: SurvivalTree(random_state=0)),
+        ("NaiveSurvivalSVM", lambda: NaiveSurvivalSVM(random_state=0, max_iter=200)),
+        ("FastSurvivalSVM", lambda: FastSurvivalSVM(random_state=0, max_iter=50)),
+        ("FastKernelSurvivalSVM", lambda: FastKernelSurvivalSVM(random_state=0, max_iter=50)),
+        ("MinlipSurvivalAnalysis", lambda: MinlipSurvivalAnalysis(solver="ecos")),
+        ("HingeLossSurvivalSVM", lambda: HingeLossSurvivalSVM(solver="ecos")),
+    ]
```
sebp, line 85: Instead of listing all estimators explicitly, have a look at `test_pandas_inputs.py` and `all_survival_estimators` from `sksurv.testing`

tests/test_polars_estimators.py
```diff
@@ -0,0 +1,514 @@
...
+    ESTIMATORS = _make_survival_estimator_constructors()
+
+    @staticmethod
+    @pytest.mark.parametrize("name,ctor", ESTIMATORS, ids=[t[0] for t in ESTIMATORS])
+    def test_estimator_polars_matches_pandas(name, ctor, whas500_encoded_small):
+        X_pd, X_pl, y = whas500_encoded_small
+        import warnings
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
sebp: Can the filter be more specific, e.g. only ignoring convergence warnings?

Applies to all instances in the tests.

tests/test_preprocessing.py
```diff
@@ -11,59 +10,24 @@
...
-            for cat in col.cat.categories[1:]:
-                name = f"{nam}={cat}"
-                s = pd.Series(col == cat, dtype=np.float64)
-                expected.append((name, s))
-        else:
-            expected.append((nam, col))
-
-    expected_data = pd.DataFrame.from_dict(OrderedDict(expected))
-    return expected_data
+    return expected_one_hot_data(data)
```
sebp: Use `expected_one_hot_data()` directly

sebp (changes requested):
Sorry it took so long for a reply. I appreciate your patience.

There are no major issues at this point, just small changes to streamline the code a little bit more. Test coverage looks excellent too.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. I replaced the boolean dtype check with `dt.is_boolean()`.

sksurv/_dataframe/_categorical_encoding.py
55Kamiryo: Done. I removed the explicit `None` return.

sksurv/_dataframe/_categorical_semantics.py
55Kamiryo: Done. I now use `dt.is_boolean()` there.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. I switched this check to `dtype.is_boolean()`.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. `get_one_hot_column_names()` now returns a list, so the extra `list(...)` conversion is no longer needed.

sksurv/_dataframe/_categorical_encoding.py
55Kamiryo: Done as part of the same change above

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. I now use `dtype.is_float()` for the float-specific NaN handling.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. I replaced that boolean dtype check with `dt.is_boolean()`.

sksurv/_dataframe/_input.py
55Kamiryo: Done. I added `# pragma: no cover` to the unreachable Oxford-join branch.

sksurv/_dataframe/_input.py
55Kamiryo: Done. I removed the redundant Narwhals dataframe wrapper and now call `nw.from_native()` directly after the eager-frame boundary check.

sksurv/datasets/base.py
55Kamiryo: Done. I now precompute both `train_col_set` and `test_col_set` before comparing them.

sksurv/datasets/base.py
55Kamiryo: Done. I removed the duplicate validation from the dataset loader wrappers and let `loadarff()` handle invalid `output_type` values.

sksurv/io/arffread.py
55Kamiryo: Done. The branches now set `values` and `dtype`, and the `pl.Series(...)` append happens once after the branch.

sksurv/kernels/_clinical_dataframe.py
55Kamiryo: Done. I now use the dict returned by `_normalize_ordinal_categories()` directly.

sksurv/kernels/clinical.py
55Kamiryo: Done. I restored the comment.

sksurv/kernels/clinical.py
55Kamiryo: Done. I clarified the message to say that `prepare(X)` sets `self.X_fit_`, which should then be passed to `fit()`.

sksurv/kernels/clinical.py
55Kamiryo: Done. I combined the dataframe-library check into a single `if` block.

sksurv/kernels/clinical.py
55Kamiryo: Done. I switched this to `hasattr`.

tests/test_clinical_kernel.py
55Kamiryo: Done. The fixture now returns `make_clinical_kernel_pandas_data` directly.

tests/test_polars_estimators.py
55Kamiryo: Done. I now build the estimator list from `all_survival_estimators()` and only apply per-estimator parameter tweaks where needed.

tests/test_polars_estimators.py
55Kamiryo: Done. I removed the blanket warning filter instead of suppressing all warnings.

tests/test_preprocessing.py
55Kamiryo: Done. The preprocessing tests now use the shared one-hot expected-data helper directly.

sksurv/_dataframe/_categorical_encoding.py
55Kamiryo: I checked this and kept the detach/reattach path.

Narwhals preserves the pandas index, and horizontal concatenation with pandas-like backends aligns by index labels. Some rebuilt columns come from NumPy and therefore have a fresh `RangeIndex`, while passthrough columns keep the caller's original index. Without resetting before concatenation, non-default or non-unique pandas indices can misalign rows.

I added documentation around this helper to make that reason explicit.

sksurv/_dataframe/_categorical_semantics.py
55Kamiryo: Done. I extracted the repeated dtype predicate into `is_categorical_or_string_dtype()` and reused it across the dataframe, preprocessing, and clinical-kernel paths.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Done. Boolean and numeric one-hot categories now stay as native values instead of being converted to strings. The generated one-hot column names are still stringified at the naming boundary.

sksurv/io/arffwrite.py
55Kamiryo: It does not make that branch obsolete. The object cast happens inside `_write_data()`, after the ARFF header has already been written.

`_write_attribute_categorical()` is called while writing the header and still sees the original categorical dtype. The object cast only applies to row values in the data section, so categorical-only frames can pass through `_check_str_array()`.

tests/test_io.py
55Kamiryo: This test covers `_to_polars_dataframe()` directly rather than the public `loadarff()` path because SciPy's public ARFF loader rejects string attributes.

The converter still needs to handle SciPy-style record arrays for non-nominal fields, so this pins the internal conversion behavior for byte-string values and the ARFF missing token `?`: bytes are decoded to strings, while `?` becomes null in the polars output.

55Kamiryo:
@sebp 
Thank you so much for the thorough review — no worries about the timing at all, and I really appreciate the concrete suggestions.

I've addressed the remaining comments and pushed the follow-up cleanup commits, and replied inline on each thread as well. The main changes:

- replaced remaining dtype `isinstance(...)` checks with Narwhals dtype predicates where applicable;
- consolidated categorical/string dtype handling;
- removed redundant dataframe wrappers/conversions;
- clarified the pandas-index detach/reattach path (kept it, with docs on why the reset is needed);
- cleaned up the estimator and helper tests;
- kept numeric/boolean one-hot categories as native values.

I reran the targeted checks, all green:

- `pytest tests/test_polars_column.py::TestEncodeCategoricalExplicitColumnsParity tests/test_datasets.py tests/test_clinical_kernel.py -q`
- `black --check sksurv/_dataframe/_column_impl.py sksurv/datasets/base.py sksurv/kernels/clinical.py`
- `ruff check sksurv/_dataframe/_column_impl.py sksurv/datasets/base.py sksurv/kernels/clinical.py`

sksurv/io/arffwrite.py
sebp: You are right, I traced the wrong path.

Is there a test that covers the "categorical-only frame" case you refer to?

tests/test_polars_estimators.py
```diff
@@ -70,45 +71,53 @@ def test_pandas_internal_container_preserved(survival_smoke_data):
...
+            if name == "CoxnetSurvivalAnalysis":
+                estimator.set_params(n_alphas=5)
+            elif name == "IPCRidge":
+                estimator.set_params(alpha=1.0)
+            elif name == "NaiveSurvivalSVM":
+                estimator.set_params(max_iter=200)
+            elif name in {"FastSurvivalSVM", "FastKernelSurvivalSVM"}:
+                estimator.set_params(max_iter=50)
+            elif name in {"MinlipSurvivalAnalysis", "HingeLossSurvivalSVM"}:
+                estimator.set_params(solver="ecos")
```
sebp: Instead of many `elif` statements, you can use [match](https://docs.python.org/3/tutorial/controlflow.html#match-statements) here.

tests/test_polars_estimators.py
```diff
@@ -70,45 +71,53 @@ def test_pandas_internal_container_preserved(survival_smoke_data):
...
+    "FastSurvivalSVM",
+    "HingeLossSurvivalSVM",
+    "IPCRidge",
+    "MinlipSurvivalAnalysis",
+}
+
+
+def _make_score_estimator_constructors():
+    return [(name, ctor) for name, ctor in _make_survival_estimator_constructors() if name in _SCORE_ESTIMATOR_NAMES]
+
```
sebp: Why the restriction on this particular subset? The `score` function is available for all survival estimators. You can check for equality of `score` in `test_estimator_polars_matches_pandas`

tests/test_polars_estimators.py
```diff
@@ -125,17 +134,13 @@ class TestSurvivalEstimatorPolarsParity:
...
+        est_pd = ctor()
+        est_pd.fit(X_pd, y)
+        pred_pd = np.asarray(est_pd.predict(X_pd), dtype=float)
 
-            est_pl = ctor()
-            est_pl.fit(X_pl, y)
-            pred_pl = np.asarray(est_pl.predict(X_pl), dtype=float)
+        est_pl = ctor()
+        est_pl.fit(X_pl, y)
+        pred_pl = np.asarray(est_pl.predict(X_pl), dtype=float)
```
sebp: Remove the `asarray` cast to check that predict returns the same dtype.

tests/test_polars_estimators.py
```diff
@@ -270,12 +259,9 @@ def test_concordance_index_censored_with_polars_derived_risk(whas500_pl_pd_small
...
 
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore")
-            risk_pd = CoxPHSurvivalAnalysis().fit(X_pd_enc, y).predict(X_pd_enc)
-            risk_pl = CoxPHSurvivalAnalysis().fit(X_pl_enc, y).predict(X_pl_enc)
+        risk_pd = CoxPHSurvivalAnalysis().fit(X_pd_enc, y).predict(X_pd_enc)
+        risk_pl = CoxPHSurvivalAnalysis().fit(X_pl_enc, y).predict(X_pl_enc)
         c_pd = concordance_index_censored(y["fstat"], y["lenfol"], risk_pd)
         c_pl = concordance_index_censored(y["fstat"], y["lenfol"], risk_pl)
         assert c_pd == c_pl
```
sebp: If you move these lines to `test_estimator_polars_matches_pandas`, you can drop this test

tests/test_polars_estimators.py
sebp: I would combine all this with `test_estimator_polars_matches_pandas`, by checking whether the estimator has `predict_survival_function` or `predict_cumulative_hazard_function`

tests/test_polars_estimators.py
```diff
@@ -398,41 +384,16 @@ def test_gridsearchcv_polars(whas500_encoded_small):
...
-        import warnings
-
         X_pd, X_pl, y = whas500_encoded_small
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore")
-            s_pd = ctor().fit(X_pd, y).score(X_pd, y)
-            s_pl = ctor().fit(X_pl, y).score(X_pl, y)
+        s_pd = ctor().fit(X_pd, y).score(X_pd, y)
+        s_pl = ctor().fit(X_pl, y).score(X_pl, y)
         assert s_pd == s_pl, f"{name}: pandas={s_pd}, polars={s_pl}"
```
sebp: See my comment above. `score` is always available and can checked in `test_estimator_polars_matches_pandas`

sksurv/io/arffwrite.py
sebp: `to_list` already gives a list

sksurv/_dataframe/_categorical_semantics.py
```diff
@@ -47,7 +52,7 @@ def get_semantic_categories(series):
         return tuple(s.cat.get_categories().to_list())
     if isinstance(dt, nw.Categorical) and s.implementation.is_pandas_like():
         return tuple(s.cat.get_categories().to_list())
-    if isinstance(dt, (nw.Categorical, nw.String, nw.Object)):
+    if is_categorical_or_string_dtype(dt):
         return tuple(sorted(s.drop_nulls().unique().to_list()))
```
sebp: You can drop the `to_list`call

sksurv/_dataframe/_column_impl.py
```diff
@@ -105,20 +101,17 @@ def encode_categorical_narwhals(table, columns=None, allow_drop=True):
 
 def _prepare_column_for_one_hot(col):
     dtype = col.dtype
-    if isinstance(dtype, nw.Boolean):
+    if dtype.is_boolean():
         # Match pandas get_dummies(drop_first=True): False is the dropped baseline.
-        col = col.replace_strict({True: "True", False: "False"}, default=None, return_dtype=nw.String)
         present = set(col.drop_nulls().unique().to_list())
```
sebp: You can drop the `to_list()` call

sksurv/_dataframe/_column_impl.py
```diff
@@ -105,20 +101,17 @@ def encode_categorical_narwhals(table, columns=None, allow_drop=True):
...
-        # Keep "False" before "True" so drop_first drops "False" as the baseline.
-        observed = [value for value in ("False", "True") if value in present]
+        observed = [value for value in (False, True) if value in present]
         return col, ColumnSemantics(name=col.name, kind="nominal", categories=tuple(observed), ordered=False)
     if dtype.is_numeric():
         ordered_values = sorted(col.drop_nulls().unique().to_list())
```
sebp: You can drop the `to_list()` call

sebp (changes requested):
Great work. Thanks a lot for the quick turnaround. The tests can be simplified further I believe. Other than that, the other issues are minor.

I noticed that coverage of the `_apply_along_column` function dropped.

```python
def _apply_along_column(array, func1d, **kwargs):
    if isinstance(array, pd.DataFrame):
        return array.apply(func1d, **kwargs)
    return np.apply_along_axis(func1d, 0, array, **kwargs)
```

The function, called by `sksurv.column.standardize`, always receives a numpy array now, so the function can be removed and `standardize` can all `np.apply_along_axis` directly.

In addition, the function `standardize_column` has partial coverage for the [if-clause on line 43](https://github.com/55Kamiryo/scikit-survival/blob/d9bd38d69333a8fb14943f72a729c546de9996dd/sksurv/column.py#L43): `if with_std`

Not sure why this changed, though. It is and wasn't covered by tests in test_column.py, AFAICT.

tests/test_polars_estimators.py
55Kamiryo: Changed the estimator-specific cases to use `match`.

tests/test_polars_estimators.py
55Kamiryo: I overlooked this when consolidating the individual score tests. It now covers all survival estimators in the main parity test.

tests/test_polars_estimators.py
55Kamiryo: Removed the casts and added an explicit dtype equality check before comparing the prediction values.

tests/test_polars_estimators.py
55Kamiryo: Moved the concordance-index comparison into `test_estimator_polars_matches_pandas` and removed the separate test.

tests/test_polars_estimators.py
55Kamiryo: Combined these checks with `test_estimator_polars_matches_pandas` using `hasattr`, reusing the already fitted estimators.

tests/test_polars_estimators.py
55Kamiryo: Done. `score` is now checked for all survival estimators in the main parity test, and the separate score test and subset helper have been removed.

sksurv/io/arffwrite.py
55Kamiryo: Removed the redundant `to_list()` call.

sksurv/_dataframe/_categorical_semantics.py
55Kamiryo: Removed the redundant `to_list()` call.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Removed the redundant `to_list()` call.

sksurv/_dataframe/_column_impl.py
55Kamiryo: Removed the redundant `to_list()` call.

sksurv/io/arffwrite.py
55Kamiryo: Yes. `test_writearff_polars_pure_categorical` and `test_writearff_pandas_pure_categorical_index_false` cover this for both backends, including the round-trip through `writearff`.

55Kamiryo:
@sebp 
Thank you again for the thorough review. I addressed all of the comments and simplified the tests as suggested.

I also tracked down the change in `standardize_column` coverage. No test was removed: before the Narwhals change, existing pandas calls with `with_std=False` reached `standardize_column` through `_apply_along_column`. Once pandas moved to the Narwhals path, those calls no longer exercised that branch.

I removed the now-redundant helper and added an explicit NumPy `with_std=False` test. `tests/test_column.py` now gives 100% branch coverage for `sksurv/column.py`.

The estimator parity checks are now consolidated into the main parameterized test, including prediction dtype, `score`, concordance index, and the optional survival and cumulative-hazard prediction APIs.

While addressing the review, I also removed three NumPy-only passthrough tests that did not use either pandas or polars inputs and added no unique line or branch coverage. The corresponding domain tests already cover those paths. I also applied the same redundant-conversion cleanup to the adjacent categorical handling in `arffwrite.py` and `_categorical_semantics.py`.

Note: The macOS CI failures look unrelated to this PR.

sebp (changes requested):
@55Kamiryo This looks great! Thank you so much for all the work.

I applied two small changes.

~~The only thing left is to rebase onto the latest version of main to get the fix for the CI error on macos.~~

sebp (approved):

55Kamiryo:
@sebp 
Thank you for taking care of the final fixes and the rebase. I really appreciate the thorough review throughout.
I look forward to continuing to contribute to this library.

PR #593, opened 2026-05-29
MNT: bump to sklearn 1.9 and adapt to it

cakedev0:
**Checklist**

- [x] pytest passes locally, with sklearn 1.9.0rc1 installed manually.
- [x] code is well formatted
- [x] CI passes

Bump scikit-learn to 1.9.0, and update the code to adapt to some changes in scikit-learn:

- remove criterion param from `GradientBoostingSurvivalAnalysis,` see PR https://github.com/scikit-learn/scikit-learn/pull/32708 
  - I remove it abruptly, but maybe you want to follow depreciation cycles like we do in scikit-learn?
- stop handling missing values in the `LogrankCriterion`, this is now handled entirely outside of criterion logic in scikit-learn. See PR https://github.com/scikit-learn/scikit-learn/pull/32119/
- Stop importing `DTYPE` from `sklearn.tree._tree`, use `np.float32` directly instead (which reflects what was done in scikit-learn)
- sample weights: call `_check_sample_weight` to align on scikit-learn
- sample weights in forests: pass them to scikit-learn bootstrap / OOB helpers, and adjust `max_samples` validation accordingly
- tests: update expected scores, I think these changes are mostly due to slight numerical behavior changes leading to different tie breaking. But I haven't check/bisect thoroughly, it would take a looot of time, even for Codex.
- tests: use `CalibratedClassifierCV` around `SVC` because `SVC(probability=True)` is now deprecated


doc/user_guide/boosting.ipynb
```diff
@@ -94,7 +94,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "0.756\n"
+      "0.765\n"
```
cakedev0, line 97: It seems the score improved! :tada: 

But according to Codex:

> I’d frame this as a numerically fragile tie exposed by sklearn tree splitter changes, not a meaningful model improvement and not directly the friedman_mse GB bugfix.

The difference only happens at iteration 88 of the GB due to an improvement delta of 3.6e-15 between two options. This is pure chance.

sksurv/ensemble/forest.py
```diff
@@ -28,6 +29,11 @@
 __all__ = ["RandomSurvivalForest", "ExtraSurvivalTrees"]
 
 MAX_INT = np.iinfo(np.int32).max
+MAX_SAMPLES_CONSTRAINTS = [
```
sebp, line 32: BaseForest already defines the constraint, but with the second element having `closed="neither"`.

see https://github.com/scikit-learn/scikit-learn/blob/cc50648cc1b759b53a4edbce0f3bb6c237349448/sklearn/ensemble/_forest.py#L198-L202

sksurv/ensemble/boosting.py
sebp: These lines should be using `csc_array` and `csr_array` according to https://github.com/scikit-learn/scikit-learn/pull/33272

sksurv/ensemble/forest.py
cakedev0: Oops, fixed.

sksurv/ensemble/boosting.py
cakedev0: Indeed, missed this one.

cakedev0:
@sebp I'm not sure to have time to spend here in the coming weeks, feel free to take over this PR.

sebp:
@cakedev0 Thanks so for much all your effort. Your changes have been merged.

PR #612, opened 2026-07-07
TESTS: consolidate test_polars_*.py into the corresponding module test files

### LINKED ISSUE #611: ENH: Consolidate test_polars_*.py into the corresponding module test files
### Feature Type

- [ ] Adding new functionality to scikit-survival

- [x] Changing existing functionality in scikit-survival

- [ ] Removing existing functionality in scikit-survival

### Problem Description

Follows up on #607.

After #592, each module that gained polars support has two test files: the original (e.g. `test_column.py`) and a polars counterpart (`test_polars_column.py`). Six such `test_polars_*.py` files now exist. This duplicates fixtures and structure, and means a future backend would add yet another parallel file rather than extending the existing tests.

### Feature Description

Fold the `test_polars_*.py` files into their module counterparts, one commit per file so each step reviews as a move rather than a rewrite.

Where a test is genuinely the same across backends, merge it into a single test parametrized over the dataframe library. A shared `dataframe_backend` fixture (in `conftest.py`) drives this; its pandas entries mirror `get_pandas_infer_string_context()`, so pandas keeps running once per option context and polars once — the number of test invocations is unchanged on every pandas version.

Everything else moves over unchanged. Most of the polars suite turned out not to mirror the pandas tests but to check polars-specific behavior (explicit `ordinal_categories`, container-type preservation, `Enum` semantics,
`LazyFrame` rejection), so those stay as dedicated tests. `test_polars_estimators.py` is kept as-is, since its tests compare pandas and polars within the same test and have no pandas counterpart to merge into.

For every consolidated file the collected test count and the coverage report are identical before and after (verified per commit); the full suite passes. This lands before the other follow-up items so that a new backend (or a new
parametrized test) extends one file per module instead of adding a parallel one.

### References and existing implementations
- #607
- #592

55Kamiryo:
#### Checklist

- [x] closes #611
- [x] pytest passes
- [x] tests are included
- [x] code is well formatted

#### What does this implement/fix? Explain your changes
This consolidates each module’s tests into a single file by folding the `test_polars_*.py` files into their corresponding test modules. It then deduplicates the mirrored pandas and Polars cases into backend-parametrized tests, as requested during review.

Each test dataset is now defined once as plain Python data:

* A backend-specific builder in `sksurv/testing/_dataframe.py` converts the data into a native pandas or Polars frame.
* The builder supports explicitly declared categorical columns using `pd.Categorical` or `pl.Enum`.
* It also exposes the appropriate assertion helpers and exception types for each backend.
* `tests/test_dataframe_backends.py` defines and verifies the builder’s behavior.

Two fixtures provide the builder:

* `dataframe_backend` runs each test once with pandas and once with Polars.
* `dataframe_backend_with_pandas_options` additionally covers the pandas `infer_string` option matrix, preserving the previous execution count for each pandas option context.

Tests that validate backend-specific behavior remain dedicated. These include:

* `LazyFrame` rejection
* `Enum` semantics
* null-versus-NaN behavior
* index preservation

`test_pandas_inputs.py` has been renamed to `test_frame_inputs.py`, and its contracts now run against both backends.

`test_polars_estimators.py` has been renamed to `test_estimator_parity.py`. It remains a separate parity test because each test directly compares the two backends.

The consolidated comparisons now use exact equality for both backends. The `1e-9` tolerance introduced in #592 was not necessary.

One existing `sksurv/` behavior has changed, as described in the PR discussion: the pandas ARFF string-attribute branch now treats `'?'` as missing, matching the Polars branch. This branch is not reachable through the public `loadarff` API.

#### AI Disclosure

I used Claude Code for this PR, under my direction and review:
- Code generation
  - this change is mostly a mechanical relocation of tests, so I had the AI implement it — one file move per commit, as I directed and I reviewed each commit.
- Text
  - as I am not a native English speaker, this description was drafted with AI assistance based on my intent, and edited by me.


tests/conftest.py
```diff
@@ -24,6 +25,33 @@ def pytest_configure(config):
...
+
+    The pandas entries mirror ``get_pandas_infer_string_context()``, so pandas
+    runs once per option context while polars runs once.
+    """
+    backend, pandas_context = request.param
+    if backend == "pandas":
+        with pandas_context:
+            yield backend
+    else:
+        yield backend
```
sebp: You can pass `contextlib.nullcontext` in `_dataframe_backend_params` then you don't need the if-else check

tests/dataframe_test_utils.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
```
sebp: I think this file should be moved to `sksurv.testing`. Potentially, converting it into directory where this file and the existing `testing.py` (renamed to base or common) resides.

tests/test_column.py
```diff
@@ -378,3 +384,535 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+                [-0.825723, -0.825723, -0.825723, -0.825723, -0.825723],
+                [-0.495434, -0.495434, -0.495434, -0.495434, -0.495434],
+                [-0.165145, -0.165145, -0.165145, -0.165145, -0.165145],
+                [0.165145, 0.165145, 0.165145, 0.165145, 0.165145],
+                [0.495434, 0.495434, 0.495434, 0.495434, 0.495434],
+                [0.825723, 0.825723, 0.825723, 0.825723, 0.825723],
+                [1.156012, 1.156012, 1.156012, 1.156012, 1.156012],
+                [1.486301, 1.486301, 1.486301, 1.486301, 1.486301],
+            ]
+        )
```
sebp: This is a exact copy of definitions in `StandardizeCase`

sebp (changes requested):
Thanks for kicking this off. It's a good start, but there still many instances were inputs being tested are duplicated. It might not be feasible to get of all duplication, but there are some obvious instances that can be consolidated.

Several options come to my mind:

Instead of creating duplicate test classes, one for pandas and one for polars, I believe it should be possible to use `pytest_generate_tests` similar to [this example](https://docs.pytest.org/en/stable/example/parametrize.html#a-quick-port-of-testscenarios). Could be useful for tests taking factory methods.

If the test data is created in a `FixtureParameterFactory` subclass, you could extract a base class that returns data as dict and two sub-classes that convert the data to data frames. If something returns only numpy array,s you don't need to overwrite the test function.

If the data frame is created within the test itself and no backend-specific magic is required, `pytest.mark.parametrize` could potentially be used by passing `pd.DataFrame` or `pl.DataFrame` as the param.

These are just my thoughts. It might make things more complicated then necessary. Let me know what you think.

55Kamiryo:
Apologies for the delayed response, and thank you for the thorough review.

On the overall direction: agreed. Consolidating the duplicated pandas/polars tests into single parametrized tests is the right goal. This is also the follow-up work that the description of #592 had deferred, so I will rework the PR in that direction rather than keep the tests in their current moved form.

After trying the three proposed mechanisms against the suite, my preference would be to adopt the second as proposed and use modified versions of the other two:

1. **`pytest_generate_tests`**

   I would prefer parametrized fixtures here. A fixture can provide the same pandas/polars matrix without introducing a collection hook, and it can also carry the pandas option contexts. This keeps the suite on a single parametrization mechanism.

   That said, switching to the hook would be straightforward if you have a strong preference for it.

2. **Base class returning dictionaries, with two converting subclasses**

   I agree with the core idea: each dataset should be defined once as plain Python data.

   Instead of adding two subclasses for each family of cases, I would inject a small backend-specific frame builder into the case methods. This produces the same result with less class boilerplate, while ensuring that nothing converts a pandas object into a polars object, or vice versa.

3. **Passing ****`pd.DataFrame`**** / ****`pl.DataFrame`**** as a parameter**

   This works well where direct construction is sufficient. The frame-builder approach above extends this idea to declared categorical types (`pd.Categorical` / `pl.Enum`), as well as each library’s assertion helpers and exception types.

   The parity tests compare `pd_out == pl_out` within the same test, so the backend cannot be a parameter there. They would continue to construct both frames from the same plain-Python literal.

Tests that are meaningful only for one backend—such as LazyFrame rejection, null-versus-NaN behavior, and the `pl.Categorical` string cache—would remain explicit single-backend tests.

The three inline comments are all clear, and I will apply them as suggested.

I have a working version of this rework locally, with the full test suite passing. Once you are comfortable with this direction, I will push the changes and document the implementation details in the PR.

tests/conftest.py
55Kamiryo: Done.
I split this into two fixtures (`dataframe_backend` for the plain pandas/polars pair, `dataframe_backend_with_pandas_options` for the `infer_string` matrix);
the polars entry passes `nullcontext()` and the if-else is gone

tests/dataframe_test_utils.py
55Kamiryo: Done.
 `sksurv/testing/` is a package now; the old `testing.py` became `_common.py`, the frame builders live in `_dataframe.py`, and`__init__.py` re-exports the old names. 

I picked underscore names over `common.py` because these helpers are test-only, and the new numpydoc
lint skips `_`-prefixed files.

tests/test_column.py
55Kamiryo: Removed.

55Kamiryo:
@sebp 
Thanks for your patience. I’ve completed the rework as discussed:

* Each dataset is defined once as plain Python data, and a backend-specific builder in `sksurv/testing/_dataframe.py` constructs the corresponding native frame. Its behavior is covered by `tests/test_dataframe_backends.py`.
* The builder is provided through the `dataframe_backend` and `dataframe_backend_with_pandas_options` fixtures.
* The mirrored tests in util, datasets, column handling, clinical kernels, preprocessing, and I/O have been consolidated.
* `test_pandas_inputs.py` has been renamed to `test_frame_inputs.py`, and its contracts now run against both backends.
* `test_polars_estimators.py` has been renamed to `test_estimator_parity.py`.
* Backend-specific tests remain separate.

### Test count and comparison strictness

I also updated the PR description. In #611, I stated that the collected test count and coverage would remain unchanged per commit, but that no longer holds after deduplication: the reviewed tip collected 1,216 tests, while the current version collects 1,270.

The consolidated comparisons are stricter as well. The `1e-9` tolerance introduced in #592 for one-hot encoding and `encode_categorical` turned out to be unnecessary, since the Polars results are bit-exact in those cases. The tests now use exact comparisons. If CI reveals a platform-specific difference, I’ll restore the tolerance.

### ARFF string-attribute behavior

One existing behavior changed. Sharing the converter test exposed inconsistent handling of `'?'` in the two ARFF string-attribute branches:

* pandas preserved it as a literal string.
* Polars treated it as missing, as required by the ARFF specification.

I changed the pandas branch to match the Polars behavior.

Because SciPy’s `loadarff` rejects string attributes, this branch is not reachable through the public API, so the change should not be user-visible. I can move it to a separate PR if you would prefer. I’ll also propose deduplicating the converter logic in a separate issue.

### Open question

When `encode_categorical` drops every column—for example, a single-category column with `allow_drop=True`—pandas returns `(n, 0)`, while Polars returns `(0, 0)`.

I left both implementations unchanged and made the shared test assert only the behavior they agree on: that no columns remain.

I think both backends should return `(n, 0)`, since a column transformation should preserve the number of samples. If you agree, I can fix that in this PR; otherwise, I’ll open a separate issue.

tests/test_clinical_kernel.py
```diff
@@ -1,59 +1,181 @@
...
+                data[name] = pd.Categorical(values, categories=declared, ordered=True)
+    if with_nominal:
+        data["metastasis"] = ["yes", "no", "yes", "yes", "no"]
+        categories["metastasis"] = ["no", "yes"]
+
+    frame = backend.make_frame(data, categories=categories)
+    expected = make_clinical_kernel_expected(
+        with_ordinal=with_ordinal, with_nominal=with_nominal, with_continuous=with_continuous
+    )
+    return frame, expected, kwargs
```
sebp: The `kwargs` are a little bit opaque, it's hard to follow that is being tested. Since this is only relates to `ordinal_categories`, can we make this explicit and return `None` for pandas and a dict for polars?

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,379 @@ def test_object_column_treated_as_nominal():
...
+        df = pl.DataFrame(
+            {
+                "age": [40.0, 50.0, 60.0, 70.0],
+                "grade": pl.Series(["I", "II", "III", "I"], dtype=pl.Enum(["I", "II", "III", "IV"])),
+            }
+        )
+        with pytest.raises(TypeError, match=r"polars\.LazyFrame is not supported"):
+            ClinicalKernelTransform().fit(df.lazy())
+        with pytest.raises(TypeError, match=r"polars\.LazyFrame is not supported"):
+            ClinicalKernelTransform(fit_once=True).prepare(df.lazy())
```
sebp: Is this test supposed to check the `transform` method after a successful fit, because currently it does not?

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,379 @@ def test_object_column_treated_as_nominal():
...
+                ),
+            }
+        )
+
+    def test_default_is_nominal(self):
+        df_pl = self._make_pl_df()
+        df_pd_ordered = self._make_pd_ordered()
+        K_pl = clinical_kernel(df_pl)
+        K_pd_ord = clinical_kernel(df_pd_ordered)
+        assert not np.allclose(K_pl, K_pd_ord)
```
sebp: This is a very generic assertion that could fail for various reasons. Best would be to compare the columns individually ("age" should match).

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,379 @@ def test_object_column_treated_as_nominal():
...
+        df_pl = pl.DataFrame(
+            {
+                "age": pl.Series("age", [40.0, 50.0, 60.0], dtype=pl.Float64),
+                "label": pl.Series("label", ["x", "y", "z"], dtype=pl.Categorical),
+            }
+        )
+        K_default = clinical_kernel(df_pl)
+        K_ordinal = clinical_kernel(df_pl, ordinal_categories={"label": ["x", "y", "z"]})
+        assert K_ordinal.shape == (3, 3)
+        assert not np.allclose(K_default, K_ordinal)
```
sebp: Same as above, it should check columns individually

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,379 @@ def test_object_column_treated_as_nominal():
...
+        out_pl = OneHotEncoder().fit_transform(df_pl).to_numpy()
+        np.testing.assert_array_equal(out_pd, out_pl, strict=True)
+
+    @staticmethod
+    def test_clinical_kernel_null_parity():
+        df_pd, df_pl = TestNominalNullParity._frames_with_null()
+        K_pd = clinical_kernel(df_pd)
+        K_pl = clinical_kernel(df_pl)
+        np.testing.assert_allclose(K_pd, K_pl, atol=1e-12, strict=True)
+        assert K_pl[1, 1] < 1.0
```
sebp: Why this assertion specifically for `K_pl`?

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,379 @@ def test_object_column_treated_as_nominal():
...
+        assert np.isnan(t_pl.X_fit_[1, 1])
+
+    @staticmethod
+    def test_object_null_parity():
+        df_pd = pd.DataFrame({"age": [40.0, 50.0, 60.0], "grade": ["I", None, "II"]})
+        df_pl = pl.DataFrame({"age": [40.0, 50.0, 60.0], "grade": ["I", None, "II"]})
+        K_pd = clinical_kernel(df_pd)
+        K_pl = clinical_kernel(df_pl)
+        np.testing.assert_allclose(K_pd, K_pl, atol=1e-12)
+        assert K_pd[1, 1] < 1.0
```
sebp: Why this assertion specifically for `K_pl`?

tests/test_column.py
```diff
@@ -158,49 +209,6 @@ def binary_with_missing(self):
         ]
         return inputs, expected
```
sebp: This class is not needed anymore, both functions are only called once by `EncodeCategoricalCases`.

tests/test_column.py
```diff
@@ -378,3 +405,407 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+        encoded = column.encode_categorical(df)
+        size_cols = [c for c in encoded.columns if c.startswith("size=")]
+        answer_cols = [c for c in encoded.columns if c.startswith("answer=")]
+        assert len(size_cols) == 3, f"expected 3 size= columns, got {size_cols}"
+        assert len(answer_cols) == 1, f"expected 1 answer= column, got {answer_cols}"
+        for c in size_cols:
+            assert c.split("=", 1)[1] in {"medium", "small", "large", "xlarge"}
+        for c in answer_cols:
+            assert c.split("=", 1)[1] in {"yes", "no"}
+
```
sebp: Can we do `assert encoded.columns == ["size=x", …]` directly, rather then doing splitting column names?

tests/test_column.py
```diff
@@ -378,3 +405,407 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+        from sksurv.column import encode_categorical
+
+        values = ["banana", "apple", "cherry", "apple", "banana"]
+        df_pd = pd.DataFrame({"fruit": pd.Categorical(values)})
+        df_pl = pl.DataFrame({"fruit": pl.Series(values, dtype=pl.Categorical)})
+        out_pd = encode_categorical(df_pd).to_numpy()
+        out_pl = encode_categorical(df_pl).to_numpy()
+        np.testing.assert_array_equal(out_pd, out_pl, strict=True)
+
```
sebp: Is identical to `test_categorical_to_numeric_pl_categorical_matches_pandas`

tests/test_column.py
```diff
@@ -378,3 +405,407 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+
+class TestLazyFramePaths:
```
sebp, line 749: Misses checking `OneHotEncoder`

tests/test_io.py
```diff
@@ -174,88 +71,151 @@ def test_writearff(make_data_fn, temp_file, infer_string_context):
...
+        # ms resolution: polars accepts only ms/us/ns datetime64 values
+        values = np.array(["2014-10-31 14:13:01", "2004-03-13 19:49:31", "1998-12-06 09:10:11"], dtype="datetime64[ms]")
+        data = backend.make_frame({"attr_datetime": values})
+        return data, "test_datetime", EXPECTED_DATETIME.copy()
 
-    enc_pd = OneHotEncoder().fit(df_pd)
-    enc_pl = OneHotEncoder().fit(df_pl)
 
-    assert list(enc_pd.categories_["grade"]) == list(enc_pl.categories_["grade"])
-    assert list(enc_pd.encoded_columns_) == list(enc_pl.encoded_columns_)
```
sebp, line 225: I believe this test related to OneHotEncoder has been lost in the refactoring. Should probably be included in test_preprocessing.py

tests/test_io.py
```diff
@@ -278,73 +239,48 @@ def test_loadarff_invalid_output_type():
...
-    """
-    with StringIO(ARFF_WITH_UNSEEN_NOMINAL) as fp:
-        df_pl = loadarff(fp, output_type="polars")
-
-    writearff(df_pl, temp_file, relation_name="round_trip", index=False)
-    temp_file.close()
-
-    with open(temp_file.name) as fp:
-        contents = fp.read()
-    # Declared categories (incl. unseen "IV") must appear in the header
```
sebp, line 294: Please preserve this comment in `test_writearff_round_trip`

tests/test_util.py
```diff
@@ -153,86 +154,109 @@ def test_from_arrays(args, kwargs, expected, expected_error):
...
+            backend, event_name="death", time_name="time_to_death", event_dtype=bool
+        )
+        return ("death", "time_to_death", data), expected, does_not_raise()
 
-    def data_no_str_columns(self):
-        data, expected = self.get_surv_data_frame(event_name=0, time_name=1)
-        inputs = (0, 1, data)
-        return inputs, expected, does_not_raise()
+    def data_no_such_column_event(self, backend):
+        data, _ = self.get_surv_data_frame(backend, event_dtype=bool)
```
sebp, line 188: This test has been lost

tests/test_util.py
```diff
@@ -153,86 +154,109 @@ def test_from_arrays(args, kwargs, expected, expected_error):
...
             TypeError,
-            match=(r"expected pandas\.DataFrame or polars\.DataFrame, " r"but got <class 'numpy.ndarray'>"),
+            match=r"expected pandas\.DataFrame or polars\.DataFrame, but got <class 'numpy.ndarray'>",
         )
-        inputs = ("event", "time", data.to_numpy())
-        return inputs, None, err
+        return ("event", "time", np.column_stack((event, time))), err
 
+    def data_polars_series(self):
+        err = pytest.raises(TypeError)
```
sebp: `pytest.raises` is missing a `match` argument

tests/test_dataframe_backends.py
```diff
@@ -0,0 +1,85 @@
...
+
+        assert isinstance(s, pd.Series)
+        assert s.name == "c"
+        assert s.dtype == pd.CategoricalDtype(["x", "y"], ordered=False)
+        assert s.isna().tolist() == [False, False, True]
+
+    @staticmethod
+    def test_missing_column_error_matches_lookup():
+        df = PANDAS_BACKEND.make_frame({"a": [1]})
+        with pytest.raises(PANDAS_BACKEND.missing_column_error):
```
sebp: `pytest.raises` has no `match` argument

tests/test_dataframe_backends.py
```diff
@@ -0,0 +1,85 @@
...
+
+        assert isinstance(s, pl.Series)
+        assert s.name == "c"
+        assert s.dtype == pl.Enum(["x", "y"])
+        assert s.null_count() == 1
+
+    @staticmethod
+    def test_missing_column_error_matches_lookup():
+        df = POLARS_BACKEND.make_frame({"a": [1]})
+        with pytest.raises(POLARS_BACKEND.missing_column_error):
```
sebp: `pytest.raises` has no `match` argument

sebp (changes requested):
Thank you, this is a huge improvement!

I believe a couple of tests have been lost in the refactor, otherwise my comments are merely about cosmetic changes.

tests/test_column.py
```diff
@@ -378,3 +405,407 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+    """Polars ``pl.Categorical`` categories must be column-scoped."""
+
+    @staticmethod
+    def test_infer_isolates_categories_per_column():
+        df = pl.DataFrame(
+            {
+                "size": pl.Series(["medium", "small", "large", "xlarge", "small"], dtype=pl.Categorical),
+                "answer": pl.Series(["yes", "no", "yes", "yes", "no"], dtype=pl.Categorical),
+            }
+        )
```
sebp: Code is replicated 3 times.

tests/test_column.py
```diff
@@ -378,3 +405,407 @@ def test_categorical_to_numeric(make_data_fn, infer_string_context):
...
+
+class TestLazyFramePaths:
+    @staticmethod
+    def _eager_lazy_pair():
+        df = pl.DataFrame(
+            {
+                "age": [40.0, 50.0, 60.0, 70.0],
+                "grade": pl.Series(["I", "II", "III", "I"], dtype=pl.Enum(["I", "II", "III", "IV"])),
+            }
+        )
```
sebp: Code is replicated 5 times across test_column.py, test_clinical_kernel.py, test_preprocessing.py. Create a fixture instead

tests/test_clinical_kernel.py
55Kamiryo: Done. `make_clinical_kernel_data` now returns `ordinal_categories` itself (None for pandas, a mapping for polars), and the call sites pass it as an explicit keyword.

tests/test_clinical_kernel.py
55Kamiryo: Right. It tested fit and prepare only. I split it: `test_fit_and_prepare_lazyframe_rejected` keeps those two, and
`test_transform_lazyframe_rejected` fits on an eager frame and asserts that transform rejects the lazy one.

tests/test_clinical_kernel.py
55Kamiryo: Done. The test now checks each column on its own. The age kernels match across the two frames, and I pinned each stage kernel to its exact matrix: equality for the nominal side, the graded ordinal kernel for the pandas side.

tests/test_clinical_kernel.py
55Kamiryo: Done. The test checks the label kernel on its own, and I pinned both results to their exact matrices: the identity for the nominal default, the graded ordinal kernel for the opt-in.

tests/test_clinical_kernel.py
55Kamiryo: The choice was arbitrary, and I should have anchored it on the reference side. Both tests now assert on `K_pd` (the parity assertion above carries the result over to `K_pl`) and pin the exact value: the age kernel (1) and the null-grade kernel (0) average to 0.5.

tests/test_clinical_kernel.py
55Kamiryo: Same fix as above: unified on `K_pd` and pinned to 0.5.

tests/test_column.py
55Kamiryo: Done. I moved both members into `EncodeCategoricalCases` and deleted `CategoricalCases`.

tests/test_column.py
55Kamiryo: Done. The test now pins the full column list in order.

tests/test_column.py
55Kamiryo: They differ in the function under test (`categorical_to_numeric` versus `encode_categorical`), so instead of deleting one I merged them into a single test parametrized over the function.

tests/test_column.py
55Kamiryo: OneHotEncoder's LazyFrame rejection lives in test_preprocessing.py (`TestOneHotEncoderLazyFrame`: fit_transform and transform); each module's tests stay in that module's file. `TestLazyFramePaths` here covers the `sksurv.column` functions.

tests/test_io.py
55Kamiryo: Restored as `test_declared_unseen_category_attrs_match_across_backends` in
test_preprocessing.py. 

I removed it as derivable: `test_loadarff_frame` pins both backends' loader output against one shared expected frame, and `TestOneHotEncoderDeclaredUnseenCategory` pins `categories_` / `encoded_columns_` on both backends, so the pandas/polars equality follows. That chain is invisible in the diff,
and an explicit test is clearer. 

The restored test builds its own frames; the loader output stays pinned in test_io.py.

tests/test_io.py
55Kamiryo: Restored.

tests/test_util.py
55Kamiryo: It moved: the case became the module-level, pandas-only `test_from_dataframe_integer_column_names` in the same file (polars requires string column names, so it cannot join the backend matrix).

I added a comment there saying so.

tests/test_util.py
55Kamiryo: Done.

tests/test_dataframe_backends.py
55Kamiryo: Done: both backends' messages contain the column name, so the match pins what the consuming tests rely on.

tests/test_dataframe_backends.py
55Kamiryo: Done.

tests/test_column.py
55Kamiryo: Done: I extracted a module-level `_size_answer_categorical_frame()` helper and use it at all three sites.

tests/test_column.py
55Kamiryo: Done: the frame is now the `polars_grade_enum_frame` fixture in conftest.py, and every site with identical content uses it. The variants with different content (the missing-value frame, for example) stay separate.

55Kamiryo:
@sebp Thank you for the review. I addressed all the comments. On the two lost tests: I restored the OneHotEncoder parity test in test_preprocessing.py, and the integer-column-names test had moved to a module-level, pandas-only test in the same file. Details are in the inline replies.

I'm also closing my own (n, 0) / (0, 0) question: I looked closer before changing anything, and there is nothing to fix. The (n, 0) case is the pandas Series path preserving the index, which is deliberate and has its own test. Every other path returns (0, 0). I'm leaving the behavior as it is.

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,398 @@ def test_object_column_treated_as_nominal():
...
+            }
+        )
+        return df_pd, df_pl
+
+    @staticmethod
+    def test_one_hot_encoder_null_parity():
+        df_pd, df_pl = TestNominalNullParity._frames_with_null()
+        out_pd = OneHotEncoder().fit_transform(df_pd).to_numpy()
+        out_pl = OneHotEncoder().fit_transform(df_pl).to_numpy()
+        np.testing.assert_array_equal(out_pd, out_pl, strict=True)
```
sebp: This test should go into test_preprocessing.py

`_frames_with_null` might need to become a fixture

tests/test_clinical_kernel.py
```diff
@@ -303,13 +432,398 @@ def test_object_column_treated_as_nominal():
...
+    @staticmethod
+    def test_clinical_kernel_null_parity():
+        df_pd, df_pl = TestNominalNullParity._frames_with_null()
+        K_pd = clinical_kernel(df_pd)
+        K_pl = clinical_kernel(df_pl)
+        np.testing.assert_allclose(K_pd, K_pl, atol=1e-12, strict=True)
+        # the non-null diagonal is exactly 1; at the null diagonal the age
+        # kernel 1 and the null grade kernel 0 average to 0.5
+        np.testing.assert_allclose(K_pd[0, 0], 1.0, atol=1e-12, strict=True)
+        np.testing.assert_allclose(K_pd[1, 1], 0.5, atol=1e-12, strict=True)
```
sebp: I would prefer checking the full diagonal:

```python
np.testing.assert_allclose(np.diag(K_pd), np.array([1.0, 0.5, 1.0]), atol=1e-12, strict=True)
```

sebp (changes requested):
@55Kamiryo excellent work!

I added two minor comments.

tests/test_clinical_kernel.py
55Kamiryo: Done. I moved the test to test_preprocessing.py (`TestOneHotEncoderNullParity`) and turned `_frames_with_null` into the `null_grade_frames` fixture in conftest.py.

tests/test_clinical_kernel.py
55Kamiryo: Done: added this line to both null-parity tests.

55Kamiryo:
@sebp Thank you for the review. Both comments are addressed.

I pushed one more commit on top. It fixes the remaining spots that had the same problems you pointed out earlier:
- exact ARFF header lines instead of substring checks in `test_io.py`
- ordered list comparisons instead of set comparisons for `encoded_columns_` and `categories_` in `test_preprocessing.py`
- one more unnecessary `list()` conversion dropped in `test_preprocessing.py`
- `strict=True` on the remaining `allclose`/`array_equal` calls
- `test_opt_in_transform` renamed to `test_opt_in_changes_fit_column_classification`, so the name says what it asserts
- small cleanups: shared test data reuse, one import hoisted, stray blank lines removed

sebp (approved):
Thanks a lot for included the last changes I requested.

